PR #1, opened 2015-04-10
Review comments for DM-2418 ("afw tests reuse the same filenames")

jdswinbank:


python/lsst/utils/tests.py
```diff
@@ -135,15 +140,56 @@ def __exit__(self, type, value, traceback):
...
+    outName = "%s_%s%s" % (callerFileName, callerFuncName, ext)
+    outPath = os.path.join(outDir, outName)
+    yield outPath
+    if os.path.isfile(outPath):
+        try:
+            os.remove(outPath)
+        except OSError as e:
+            print "Warning: could not remove file %r: %s" % (outPath, e)
+    else:
+        print "Warning: could not find file %r" % (outPath,)
```
jdswinbank: Should this be using logging rather than `print`? (That's a genuine question -- I don't know the answer, and I'm not aware of a standard -- so I ask for my information rather than as a correction.)

python/lsst/utils/tests.py
```diff
@@ -135,15 +140,56 @@ def __exit__(self, type, value, traceback):
...
+    """
+    callerFrameInfo = inspect.getframeinfo(inspect.stack()[2][0])
+    callerDir, callerFileNameWithExt = os.path.split(callerFrameInfo.filename)
+    callerFileName = os.path.splitext(callerFileNameWithExt)[0]
+    outDir = os.path.join(callerDir, ".tests")
+    if not os.path.isdir(outDir):
+        outDir = ""
+    callerFuncName = callerFrameInfo.function
+    outName = "%s_%s%s" % (callerFileName, callerFuncName, ext)
+    outPath = os.path.join(outDir, outName)
```
jdswinbank: This is almost perfect, but: I often find myself writing code where I define the basic structure of a test, then call it repeatedly using different arguments. For example:

```python
class myTestCase(unittest.TestCase):
    def _testImpl(self, *args):
        # Do some test that uses args
    def test1(self):
         self._testImpl(arg1)
    def test2(self):
         self._testImpl(arg2)
```

(For a real life example, see eg https://github.com/lsst/afw/blob/master/tests/statistics.py#L183).

I think that using the current implementation of `getTempFilePath`, any temporary files generated here would be named after `_testImpl`, which risks a collision -- thought ought to be named after `testN`.

I think you should be able to avoid this by setting `outName` to something like:

```python
outName = "%s%s" % (inspect.getargvalues(inspect.stack()[2][0]).locals['self'].id().split('.', 1)[1].replace('.', '_'), ext)
```

(obviously, reformatted to something more readable!)

python/lsst/utils/tests.py
```diff
@@ -115,12 +117,15 @@ def findFileFromRoot(ifile):
 #-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
 
 class temporaryFile:
-    """A class to be used with a with statement to ensure that a file is deleted
-E.g.
+    """!A class to be used with a with statement to ensure that a file is deleted
+
+    @warning deprecated: use getTempFilePath instead
```
jdswinbank: Is there a reason to maintain & deprecate this rather than removing it outright? I think you've removed the only place it was being used as part of this ticket, so just trashing it shouldn't break anything.

python/lsst/utils/tests.py
```diff
@@ -23,18 +23,20 @@
 """Support code for running unit tests"""
 
 import unittest
+import inspect
+from contextlib import contextmanager
 import warnings
+import os
+import sys
+import gc
```
jdswinbank: Being really picky -- the C++ coding standard says that `include`s should be sorted alphabetically. I don't think the Python standard specifically has a similar requirement, but maybe it's implied?

python/lsst/utils/tests.py
r-owen: I think print is better because this is intended for unit tests.

Logging would definitely be preferred if this was for pipeline execution.

python/lsst/utils/tests.py
r-owen: OK. I usually do use alphabetical, but didn't want to mess with existing code. I'll reorder them.

python/lsst/utils/tests.py
r-owen: I wasn't sure if it was used elsewhere (it wasn't used in any code I had checked out, but I am missing a few packages), but I agree. I'll try it and see what buildbot says.

python/lsst/utils/tests.py
r-owen: Note that the problem you mention won't cause a collision, because tests within a test file are not run in parallel. But your alternative still sounds better.

PR #5, opened 2016-04-27
Update testing infrastructure for pytest

timj:
* Add support for testing binaries.
* Reset leak counter after memory test runs.
* flake8 cleanups.

### DIFF
--- a/python/lsst/utils/tests.py
+++ b/python/lsst/utils/tests.py
@@ -27,6 +27,7 @@ from contextlib import contextmanager
 import gc
 import inspect
 import os
+import subprocess
 import sys
 import unittest
 import warnings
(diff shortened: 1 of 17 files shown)


python/lsst/utils/tests.py
```diff
@@ -61,40 +73,169 @@ def run(suite, exit=True):
...
+    def _test_builder(root_dir, test_binary):
+        """!Build a test method.
+
+        The method is built for the supplied binary located
+        in the supplied root directory. If test_binary is absolute
+        path root_dir is ignored.
+
+        test_function = cls._test_builder(root, binary)
+
+        This function does not attach the created test function to a class.
```
r-owen: I don't understand and would appreciate rewording or more explanation. I can think of (at least) two possible meanings:

- The returned function is an unbound method of this class, so you must explicitly specify the "self" argument.
- The function is a free function.

python/lsst/utils/tests.py
```diff
@@ -61,40 +73,169 @@ def run(suite, exit=True):
...
+        The returned will have a name derived from the test_binary.
```
r-owen: Please add a word after "returned", as in "the returned method", or "the name of the returned method is derived..."

timj: Yes. Sorry. Typo.

python/lsst/utils/tests.py
timj: It returns a a free function that is expected to be bound to a class. This may really be an artifact of the way I developed it. I could remove this static method completely and do everything in `discover_tests` or I could make `_test_builder` a class method without returning anything.

python/lsst/utils/tests.py
```diff
@@ -43,10 +45,20 @@
     memId0 = 0                          # ignore leaked blocks with IDs before memId0
     nleakPrintMax = 20                  # maximum number of leaked blocks to print
 
+try:
+    type(open_files)
+except NameError:
+    open_files = set([p.path for p in psutil.Process().open_files()])
```
r-owen: It would be more idiomatic and likely more efficient to remove the []

python/lsst/utils/tests.py
```diff
@@ -43,10 +45,20 @@
...
+
+
 def init():
     global memId0
+    global open_files
     if dafBase:
-        memId0 = dafBase.Citizen_getNextMemId() # used by MemoryTestCase
+        memId0 = dafBase.Citizen_getNextMemId()  # used by MemoryTestCase
+    # Reset the list of open files
+    open_files = set([p.path for p in psutil.Process().open_files()])
```
r-owen: Since you set open_files here, is there any need for the try type(open_files) above? It would be simpler and I think easier to understand (and avoid code duplication) to replace the try/except above with:

    open_files = set()

python/lsst/utils/tests.py
```diff
@@ -61,40 +73,169 @@ def run(suite, exit=True):
...
                 self.fail("Leaked %d blocks" % dafBase.Citizen_census(0, memId0))
 
-#-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
+    def testFileDescriptorLeaks(self):
+        now_open = set([p.path for p in psutil.Process().open_files()])
+        diff = now_open.difference(open_files)
+        if diff:
+            for f in diff:
+                print("File open: %s" % f)
+            self.fail("Failed to close %d files" % len(diff))
```
r-owen: This is a really nice test. It would be helpful if you would print the first few file paths, as well, e.g. `sorted(diff)[0:3]` or something similar.

timj: In the error message? All the open files are printed (as for memory leaks) and `py.test` will capture that output on failure and report it.

python/lsst/utils/tests.py
```diff
@@ -61,40 +73,169 @@ def run(suite, exit=True):
...
+                    print("...")
                     for i in range(nleakPrintMax - 1, -1, -1):
-                        print census[i].repr()
+                        print(census[i].repr())
 
                 self.fail("Leaked %d blocks" % dafBase.Citizen_census(0, memId0))
 
-#-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
+    def testFileDescriptorLeaks(self):
+        now_open = set([p.path for p in psutil.Process().open_files()])
```
timj: If we go down this path, I'll probably put the above code in a staticmethod so that `init()` and the test agree on definitions.

python/lsst/utils/tests.py
r-owen: You're right. I missed the prints above. Best to leave it as is,

python/lsst/utils/tests.py
```diff
@@ -61,40 +73,169 @@ def run(suite, exit=True):
...
+
+class BinariesTestCase(unittest.TestCase):
+    """!Test that executable binaries can be run and return good status.
+
+    The test methods are dynamically created. Callers
+    must subclass this class in their own test file and invoke
+    the discover_tests() class method to register the tests.
+    """
+
+    def assertExecutable(self, executable, args=None, root_dir=None):
```
r-owen: I suggest you add a msg argument, for the message to print if the executable fails (default to None and print something sensible).

Please standardize on "binary" or "executable" to refer to the path to an executable binary file in this module. Also, please standardize argument names and order for this method and `_test_builder`, for those arguments both have in common.

PR #7, opened 2016-05-05
Skip NFS files in fd leak test and fix grammar in failure message

timj:


python/lsst/utils/tests.py
```diff
@@ -124,13 +125,14 @@ def testFileDescriptorLeaks(self):
         now_open = _get_open_files()
 
         # Some files are opened out of the control of the stack.
-        now_open = set(f for f in now_open if not f.endswith(".car"))
+        now_open = set(f for f in now_open
+                       if not f.endswith(".car") and ".nfs" not in f)
```
ktlim: I thought `.nfs` files were left behind when a process has a file open on a network filesystem and that file is removed (unlinked) from its directory.  These files are supposed to be cleaned up after the file is closed.  If so, doesn't the presence of a `.nfs` file attached to an open file descriptor indicate that there really is an open file that shouldn't be?

timj: Over the years I've had weirdness relating to NFS that relate to issues where it takes a relatively large amount of time for NFS resources to get freed. Given that the tests all work fine on local disk without any file leakage I am inclined to put this down to NFS race conditions and keep the code in there.

PR #12, opened 2016-07-21
DM-6998: Add helper function to fix memory test ordering

timj:


timj:
@ktlim wonders if it wouldn't be easier to define `load_tests` and `setup_module` in `lsst.utils.tests` and import them into the test module:
```python
from lsst.utils.tests import load_tests, setup_module
```
`load_tests` is a well-defined item and it may be easier to say we do it that way rather than two lines of boiler plater later on. I haven't thought too hard about whether `setup_module` should be done like that as well.

PR #13, opened 2016-07-29
DM-6320: Port to Python 3

timj:


tests/raDecToStr.py
```diff
@@ -22,6 +22,7 @@
 # see <http://www.lsstcorp.org/LegalNotices/>.
 #
 
+from builtins import range
```
PaulPrice: Commit message doesn't mention `range`.

python/lsst/utils/multithreading/SharedData.py
```diff
@@ -101,7 +101,7 @@ def __getattribute__(self, name):
         return self._d[name]
 
     def __setattr__(self, name, value):
-        if name == "_d" or len(self._d) == 0 or name in self.__dict__.keys():
+        if name == "_d" or len(self._d) == 0 or name in self.__dict__:
```
PaulPrice: I think "modern python" would have `self._d` instead of `len(self._d) == 0`.

python/lsst/utils/multithreading/SharedData.py
```diff
@@ -124,19 +124,19 @@ def initData(self, data):
         """
         with self._cond:
             bad = []
-            realattrs = self.__dict__.keys()
-            for key in data.keys():
+            realattrs = list(self.__dict__.keys())
```
PaulPrice: `set` would be better because we want to see what's `in` it.

In fact, I would write the whole `for` loop that's about to come as a `set` operation:

    bad = set(data.keys()).intersection(set(self.__dict__.keys()))

python/lsst/utils/tests.py
```diff
@@ -22,6 +22,9 @@
 
 """Support code for running unit tests"""
 from __future__ import print_function
+from __future__ import division
```
PaulPrice: Is there a reason we don't put this in every file?

Any reason not to combine it with the above line (I actually like the style of one `import` per line, but that's not what we generally do).

timj: I believe coding policy is that these `__future__` imports should always be used (and absolute_import as well for init files) but retrospectively adding the division one usually needs some thinking.

python/lsst/p_lsstSwig.i
```diff
@@ -53,6 +53,12 @@
 %include "carrays.i"
 %include "typemaps.i"
 
+// Allow Python2 to accept unicode strings
+%module unicode_strings
```
PaulPrice: Why `%module`?  I don't think you want to do that --- it's declaring a new module, but this is supposed to be included by other modules.

timj: You are right. I was probably a bit too trigger happy on copying the example code from http://swig.org/Doc3.0/Python.html#Python_2_unicode (albeit this code does actually solve the problem). I'll remove everything except for the `#define` and see what happens.

python/lsst/utils/multithreading/SharedData.py
timj: I think it's `not self._d` and I've made the change.

tests/raDecToStr.py
timj: Message updated

python/lsst/p_lsstSwig.i
timj: I removed the module and I also added a test that fails without this unicode flag enabled.

PR #14, opened 2016-08-01
Wrap with pybind11 instead of swig

pschella:


python/lsst/utils/SConscript
```diff
@@ -1,4 +1,6 @@
 # -*- python -*-
 from lsst.sconsUtils import scripts
-scripts.BasicSConscript.python()
```
timj: The `python` method only built SWIG files and did nothing else?

python/lsst/utils/SConscript
```diff
@@ -1,4 +1,6 @@
...
+scripts.BasicSConscript.pybind11(['raDecStr'])
+scripts.BasicSConscript.pybind11(['demangle'])
+scripts.BasicSConscript.pybind11(['utils'])
```
timj: Can this target take a list with multiple files rather than one per line? Could it scan the directory for `.cc` files?

python/lsst/utils/utilsLib.py
```diff
@@ -0,0 +1,3 @@
+from _raDecStr import *
+from _demangle import *
+from _utils import *
```
timj: Missing new line. I assume this file is for backwards compatibility and can be removed when SWIG goes.

timj: Actually, can't we just move these 3 lines to `__init__.py`?

timj:
Having looked at the other PRs, ignore my comment about the .cc files in the python tree.

python/lsst/utils/utilsLib.py
timj: From the other PRs I see you are just being consistent and declaring that Python wrapper python code always lives in a Lib.py file just like for SWIG.

python/lsst/p_lsstSwig.i
```diff
@@ -1,281 +0,0 @@
-// -*- lsst-c++ -*-
```
pschella: It is, but the plan is to migrate the stack from the bottom up on a separate branch and then switch over in one go later. Although technically possible, keeping swig alive for higher layers while porting lower layers to pybind11 adds a lot of complexity that was deemed unnecessary (at least for the trial). I suggest that we leave it out for now, but consider restoring it if it turns out that the higher layers are otherwise independent.

python/lsst/utils/SConscript
pschella: Yes, from looking at `sconsUtils` that is all it does (despite the slightly confusing name).

python/lsst/utils/SConscript
pschella: Yes, we can give it all modules in one list instead (and I will make that change). It could certainly also scan the directory, but that would require modifying `sconsUtils` so I propose to not do that for now.

python/lsst/p_lsstSwig.i
timj: Ok. I had not realized we were going to have to "big bang" this. As you move higher up the stack there will be lots of rebasing to do.

python/lsst/utils/SConscript
timj: Fair enough.

python/lsst/utils/utilsLib.py
pschella: Yes, I originally wanted to have everything in `__init__.py`. But that turned out to require too much unnecessary churn in the unit tests. So I opted for clarity and have this two step approach (with the old `packageLib.py` in between the now separate modules and the package init).

pschella:
Thanks!

PR #16, opened 2016-08-07
DM-7150: Configure a default log for app unit tests

hchiang2:


python/lsst/utils/tests.py
```diff
@@ -71,6 +72,13 @@ def init():
         memId0 = dafBase.Citizen_getNextMemId()  # used by MemoryTestCase
     # Reset the list of open files
     open_files = _get_open_files()
+    lsst.log.configure_prop("""
```
timj: Can you add a comment before this line to explain what you are doing.

python/lsst/utils/tests.py
```diff
@@ -71,6 +72,13 @@ def init():
         memId0 = dafBase.Citizen_getNextMemId()  # used by MemoryTestCase
```
timj: I know this is not your fault, but please add a docstring to this function to explain what it does. It didn't really matter when the only thing it did was set the watermark for the memory leak tester.

hchiang2: docstring added

python/lsst/utils/tests.py
hchiang2: comment added

python/lsst/utils/tests.py
```diff
@@ -65,12 +66,21 @@ def _get_open_files():
...
 
 def init():
+    """Initialize the memory tester and configure a default log."""
     global memId0
     global open_files
     if dafBase:
         memId0 = dafBase.Citizen_getNextMemId()  # used by MemoryTestCase
     # Reset the list of open files
     open_files = _get_open_files()
+    # Setup a default cononfiguration for log used in the test framework
```
timj: Typo "cononfiguration"

hchiang2: Thanks for catching it.

PR #17, opened 2016-08-10
DM-7186: Do not fail if tests leave SSSD passwd file open.

jdswinbank:
This happens on the new (August 2016) NCSA LDAP infrastructure; it's triggered
by:

  from eups import Eups

### DIFF
--- a/python/lsst/utils/tests.py
+++ b/python/lsst/utils/tests.py
@@ -177,7 +177,8 @@ class MemoryTestCase(unittest.TestCase):
         now_open = _get_open_files()
 
         # Some files are opened out of the control of the stack.
-        now_open = set(f for f in now_open if not f.endswith(".car") and not f.endswith(".ttf"))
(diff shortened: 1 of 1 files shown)


python/lsst/utils/tests.py
```diff
@@ -177,7 +177,8 @@ def testFileDescriptorLeaks(self):
         now_open = _get_open_files()
 
         # Some files are opened out of the control of the stack.
-        now_open = set(f for f in now_open if not f.endswith(".car") and not f.endswith(".ttf"))
+        now_open = set(f for f in now_open if not f.endswith(".car") and not f.endswith(".ttf")
+                                              and not f == "/var/lib/sss/mc/passwd")
```
timj: `f  !=` ? Should we be saying `not f.endswith("/passwd")` instead?

jdswinbank: Yes, `!=` is better. I decided to make the test as specific as possible for now, since it seems like it's just this one file causing problems. We can always loosen it up to an `endswith()` later if we need to.

PR #18, opened 2016-08-18
Added equality helper and assert message for allClose

timj:


python/lsst/utils/tests.py
```diff
@@ -538,7 +538,8 @@ def plotImageDiff(lhs, rhs, bad=None, diff=None, plotFileName=None):
 
 @inTestCase
 def assertClose(testCase, lhs, rhs, rtol=sys.float_info.epsilon, atol=sys.float_info.epsilon, relTo=None,
-                printFailures=True, plotOnFailure=False, plotFileName=None, invert=False):
+                printFailures=True, plotOnFailure=False, plotFileName=None, invert=False,
+                msg=None):
```
TallJimbo: Formatting just looks a little weird; why not just add this to the previous line?  (Obviously not a big deal.)

r-owen:
I agre with @timj

python/lsst/utils/tests.py
```diff
@@ -566,6 +567,7 @@ def assertClose(testCase, lhs, rhs, rtol=sys.float_info.epsilon, atol=sys.float_
     @param[in]  invert         If True, invert the comparison and fail only if any elements *are* equal.
                                Used to implement assertNotClose, which should generally be used instead
                                for clarity.
+    @param[in] msg             String to include in error message when assert fails.
```
r-owen: The string appears to be appended; is it worth documenting that?

python/lsst/utils/tests.py
r-owen: I agree.

Also (existing issue) lhs and rhs are documented to be scalars or np arrays, but this code also works with other kinds of sequences of numbers. Please document that fact.

python/lsst/utils/tests.py
```diff
@@ -631,3 +636,13 @@ def assertNotClose(testCase, lhs, rhs, **kwds):
     See assertClose for more information.
     """
     return assertClose(testCase, lhs, rhs, invert=True, **kwds)
+
+
+@inTestCase
+def assertEqualFloatingPoint(testCase, lhs, rhs, **kwargs):
```
r-owen: I'm not convinced we need this, since assertEqual does the same job, but if you are doing to add it, I suggest a name assertFloatsEqual, i.e. put the type before the comparison, preferably abbreviated a bit), to match existing assertAngles..., assertWcs..., assertImages..., etc.

Also, please add assertFloatsNotEqual for symmetry.

tests/utils.py
```diff
@@ -42,7 +42,9 @@
 
 
 class UtilsTestCase(lsst.utils.tests.TestCase):
-    """A test case for Utils"""
+    def setUp(self):
+        self.a = numpy.zeros((5, 5), dtype=float)
+        self.b = numpy.zeros((5, 5), dtype=float)
```
r-owen: I'm a bit surprised that you changed to preallocate this, since it's so fast to allocate what you want, but it's just a matter of taste

python/lsst/utils/tests.py
timj: and also for symmetry with `assertListEqual` from `unittest`.

python/lsst/utils/tests.py
parejkoj: Good point: fixed.

python/lsst/utils/tests.py
parejkoj: I've reworked the formatting to be clear with the new name, and clarified the lhs/rhs docstrings.

python/lsst/utils/tests.py
parejkoj: unittest.assertEqual doesn't work on numpy arrays, nor on numpy array vs. scalar comparisons, hence why we need the exact equality version. I've renamed them all to assertFloats*.

tests/utils.py
parejkoj: It's not a speed question, but a "we need these everywhere, and might as well be consistent about it." It doesn't really matter, and I can move it all back into the individual functions. I've now split up them up even more, to have each test only test one method.

r-owen: Fair enough. No need to change it.

python/lsst/utils/tests.py
```diff
@@ -620,14 +623,42 @@ def assertClose(testCase, lhs, rhs, rtol=sys.float_info.epsilon, atol=sys.float_
...
+    """
+    Assert that lhs == rhs (both numeric types, whether scalar or array).
+
+    See assertClose (called with rtol=atol=0) for more information.
+    """
+    return assertFloatsAlmostEqual(testCase, lhs, rhs, rtol=0, atol=0, **kwargs)
+
+
+@inTestCase
+def assertClose(*args, **kwargs):
```
r-owen: sweet

tests/utils.py
```diff
@@ -21,79 +21,161 @@
...
-        self.assertNotClose(a, b, rtol=1E-12, relTo=100.0)
-        a[:, :] = numpy.arange(-12, 13).reshape(5, 5)
-        b[:, :] = a
-        b[2, :] += numpy.linspace(-1E-4, 1E-4, 5)
-        self.assertClose(a, b, rtol=1E-3, atol=1E-4)
-        if False:
-            # set to True to test plotting and printing by-eye when tests fail
+    def setUp(self):
+        self.large = 100.
+        self.epsilon = 1e-8
```
r-owen: Why not use sys.float_info.epsilon here? No big deal, just curious.

tests/utils.py
```diff
@@ -21,79 +21,161 @@
...
+        self.zeros = np.zeros((5, 5), dtype=float)
+        self.zeros2 = np.zeros((5, 5), dtype=float)
+        self.epsilons = np.zeros((5, 5), dtype=float) + self.epsilon
+        self.larges = np.zeros((5, 5), dtype=float) + self.large
+        self.larges2 = np.zeros((5, 5), dtype=float) + self.large
```
r-owen: Why pre-compute the copies? I suspect it would be slightly clearer to set larges2=self.larges2.copy() when needed. But no big deal.

tests/utils.py
parejkoj: This epsilon is used in some tests, while others just take the default value, to test both the defaults and passing a kwarg.

tests/utils.py
parejkoj: I don't want to test `x==x`, but rather `x==y` where y was create in the same way as x. Maybe the distinction isn't important. But this is the place to do it, so I can always just use the thing later on.

Also, the places I use `.copy()` here are so that I can slightly tweak the values for later testing.

PR #20, opened 2016-08-23
Tickets/dm 7270

parejkoj:
Renamed some tests to start with "test" and cleaned up some code clutter.


tests/testGetPackageDir.py
```diff
@@ -46,5 +47,5 @@ def setup_module(module):
     lsst.utils.tests.init()
 
 if __name__ == "__main__":
-    lsst.utils.tests.init()
+    setup_module(sys.modules[__name__])
```
timj: Other files in the stack do this so I have no problem with it happening here.

PR #21, opened 2016-09-06
DM-7137: utils raDecStr delimiters are untested and don't work

jmeyers314:
I added tests that use the delimiters, and fixed/pythonized/deduplicated the code.

### DIFF
--- a/src/RaDecStr.cc
+++ b/src/RaDecStr.cc
@@ -146,14 +146,18 @@ std::string ut::raDecDegToStr(double raDeg, double decDeg) {
 // Converting strings to numbers
 //
 
-double ut::raStrToRad(std::string raStr, std::string delimiter) {
-    return degToRad( raStrToDeg(raStr, delimiter) );
-}
-
-
-double ut::raStrToDeg(std::string raStr, std::string delimiter) {
-
-    //Regex the hours, minutes and seconds
+/**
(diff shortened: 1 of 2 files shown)


src/RaDecStr.cc
```diff
@@ -121,111 +121,93 @@ string ut::decDegToStr(double decDeg) {
...
 //
 // Converting strings to numbers
 //
 
-double ut::raStrToRad(std::string raStr, std::string delimiter) {
-    return degToRad( raStrToDeg(raStr) );
-}
-
+double sexagesimalStrToDecimal(std::string inStr, std::string delimiter) {
+    double hh_or_dd, mm, ss;
```
timj: why were these moved to the top of the function?

timj:
Do we understand why it wasn't working for me with non-standard delimiters?

src/RaDecStr.cc
jmeyers314: Good point.  I'll declare down below where they first get used.

jmeyers314:
>Do we understand why it wasn't working for me with non-standard delimiters?

I believe this was because the `boost::regex` object was initialized `static`, so it got set the first time through the function to use ':' as the delimiter (in the earlier tests), then ignored any request for a different delimiter when returning to the function in your new test.

PR #26, opened 2017-02-01
DM-8805: utilities for pybind11 wrappers

TallJimbo:
This branch adds two major pieces of utility code to aid in wrapping C++ code in downstream packages:
 - a pair of decorators for adding new attributes to existing (presumably wrapped classes)
 - a metaclass in the spirit of `collections.ABCMeta` that helps to tie together different instantiations of the same wrapped template.

While the former is really just about code readability for supplemental wrapper code, the latter provides new more Python (and backwards compatible) interfaces for the classes wrapped with it.

### DIFF
--- a/include/lsst/utils/pybind11.h
+++ b/include/lsst/utils/pybind11.h
@@ -53,7 +53,7 @@ This is wrapped as follows for `lsst::afw::table::BaseRecord`, where `cls` is an
 
     utils::addSharedPtrEquality<BaseRecord>(cls);
 
-Note that all record sublasses inherit this behavior without needing to call this function.
+Note that all record subclasses inherit this behavior without needing to call this function.
 */
 template<typename T, typename PyClass>
 inline void addSharedPtrEquality(PyClass & cls) {
--- a/python/lsst/utils/wrappers.py
+++ b/python/lsst/utils/wrappers.py
@@ -25,12 +25,9 @@ def isAttributeSafeToTransfer(name, value):
     classes, leaving only those explicitly defined in a class decorated by
     `continueClass` or registered with an instance of `TemplateMeta`.
     """
-    if name.startswith("__"):
-        if (value is not getattr(object, name, None) and
-                name not in INTRINSIC_SPECIAL_ATTRIBUTES):
-            return True
-        else:
-            return False
+    if name.startswith("__") and (value is getattr(object, name, None) or
+                                  name in INTRINSIC_SPECIAL_ATTRIBUTES):
+        return False
     return True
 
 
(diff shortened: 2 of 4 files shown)


python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,298 @@
...
+        import numpy as np
+        from ._image import ImageF, ImageD
+
+        class Image(metaclass=TemplateMeta):
+            pass
+
+        Image.register(np.float32, ImageF)
+        Image.register(np.float64, ImageD)
+        Image.alias("F", ImageF)
+        Image.alias("D", ImageD)
```
pschella: How will this work for classes that are templated on more than one type?

pschella: The answer probably is: "it doesn't". But in that case we might want to be able to just register without a `dtype`, and just have the common ABC functionality.

pschella: Don't know if that is common enough to worry about though.

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,298 @@
...
+        return type.__new__(cls, name, bases, attrs)
+
+    def __call__(self, *args, **kwds):
+        # __call__ is invoked when someone tries to construct an instance of
+        # the abstract base class.  We look a dtype kwarg and use that to
+        # look up which derived class to construct, then forward all the
+        # remaining arguments to that constructor.
+        dtype = kwds.pop("dtype", None)
+        if dtype is None:
+            raise TypeError("A dtype keyword argument must be provided.")
```
pschella: Do you want to enable a default dtype to be registered? Like in numpy.

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,298 @@
...
+        attributes will thus *replace* existing attributes in those classes
+        with the same name, and subclasses cannot delegate to base class
+        implementations of these methods.
+
+    Finally, abstract base classes that use ``TemplateMeta`` define a dict-
+    like interface for accessing their registered subclasses, providing
+    something like the C++ syntax for templates::
+
+        Image[np.float32] -> ImageF
+        Image["D"] -> ImageD
```
pschella: This still feels a bit like exposing too much internal stuff to the users. But I'm not sure what else to do either.

tests/testWrappers.py
```diff
@@ -0,0 +1,239 @@
...
+        self.Base.register(np.float64, self.DerivedD)
+
+    def alias(self):
+        self.Base.alias("F", self.DerivedF)
+        self.Base.alias("D", self.DerivedD)
+
+    def testCorrectRegistration(self):
+        self.register()
+        self.assertEqual(self.DerivedF.dtype, np.float32)
+        self.assertEqual(self.DerivedD.dtype, np.float64)
```
pschella: What happens if Derived already has a `dtype`? Will it clash?

tests/testWrappers.py
```diff
@@ -0,0 +1,239 @@
...
+            def property1(self):
+                return False
+
+        class DerivedF:
+            pass
+
+        class DerivedD:
+            pass
+
+        self.Base = Base
```
pschella: Not really a base class right? Perhaps better name them: `Example`, `ExampleF` and `ExampleD`.

python/lsst/utils/wrappers.py
TallJimbo: I was actually thinking it'd just work if we used tuples as dtype keys, though some of the analogies with numpy would obviously fall apart. It might be good to add that to the tests to see if I'm right about that.

python/lsst/utils/wrappers.py
TallJimbo: This is actually the one part of this I'm afraid we need to keep for backwards compatibility (at least if we're going to use these ABCs in the places we were using dicts before).  But I have no objection to phasing them out later, especially if we provide some other (less public) way to do these lookups.

tests/testWrappers.py
TallJimbo: Well, the `collections` builtin module would call this an abstract base class, even though it isn't a true base class.  But I think that your change is worth making anyway to make the tests look more like real-world usage would (while tests shouldn't have to serve as docs, it's not a bad thing to make them better at it if it doesn't hurt anything else).

python/lsst/utils/wrappers.py
TallJimbo: Do you have a strong opinion? I don't - I could imagine it being useful, but I don't have a specific template class in mind for that, and I could also see it being abused.

tests/testWrappers.py
TallJimbo: I think it will silently overwrite it, unless it's a property that isn't settable.  I'm thinking an exception may be better.

python/lsst/utils/wrappers.py
TallJimbo: While trying to use this with afw, I've realized it'd be nice to be able to use a variable other than `dtype`, and possibly split any tuple keys into multiple attributes (e.g. `dtype` and `dimensions`); it's a bit weird using `dtype` when the key is an integer.

TallJimbo:
I've just added a number of commits to improve support for template with multiple parameters and parameters that aren't conceptually a "dtype".  I also fixed line length issues.  Please don't worry too much about commit-separation here; I plan to squash a lot later, and have separated my commits with that in mind.

python/lsst/utils/wrappers.py
TallJimbo: I've found a use case: I think it'd be nice for `Point(4, 3, dtype=float)` to work without an explicit `dimensions=2` argument as well, so I've just added support for this.

python/lsst/utils/wrappers.py
pschella: Indeed.

python/lsst/utils/wrappers.py
pschella: Fair enough.

pschella:
Please also rebase onto latest DM-8467 branch.

include/lsst/utils/pybind11.h
```diff
@@ -0,0 +1,119 @@
...
+
+Example:
+
+lsst::afw::table records are considered equal if two `std::shared_ptr<record>` point to the same record.
+This is wrapped as follows for `lsst::afw::table::BaseRecord`, where `cls` is an instance of
+`py::class_<BaseRecord, std::shared_ptr<BaseRecord>>)`:
+
+    utils::addSharedPtrEquality<BaseRecord>(cls);
+
+Note that all record sublasses inherit this behavior without needing to call this function.
```
natelust: subclasses*

include/lsst/utils/pybind11.h
```diff
@@ -0,0 +1,119 @@
...
+
+#include "lsst/pex/exceptions.h"
+
+namespace py = pybind11;
+
+namespace lsst {
+namespace utils {
+
+/**
+Add `__eq__` and `__ne__` methods based on two std::shared_ptr<T> pointing to the same address
```
natelust: Its a shame we cannot overload is and is not

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+
+def isAttributeSafeToTransfer(name, value):
+    """Return True if an attribute is safe to monkeypatch-transfer to another
+    class.
+
+    This rejects special methods that are defined automatically for all
+    classes, leaving only those explicitly defined in a class decorated by
+    `continueClass` or registered with an instance of `TemplateMeta`.
+    """
+    if name.startswith("__"):
```
natelust: why not make this:
if name.startswith("__") and (value is getattr(object, name, None) or name in 
    INTRINSIC_SPECIAL_ATTRIBUTES):
        return False
return True

natelust: Please excuse the poor formatting, it got lost in the comment, I hope you get the meaning

TallJimbo: I found this a bit easier to read,(it avoided the wrapped if statement once up on a time, which I tend to find very hard to follow in any language).  But since it doesn't avoid it anymore, and it just made you ask why I did it, it should probably go back.

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+
+    Standard decorators like ``classmethod``, ``staticmethod``, and
+    ``property`` may be used *after* this decorator.  Custom decorators
+    may only be used if they return an object with a ``__name__`` attribute
+    or the ``name`` optional argument is provided.
+    """
+    def decorate(func):
+        # Using 'name' instead of 'name1' breaks the closure because
+        # assignment signals a strictly local variable.
+        name1 = name
```
natelust: This is another place it is a shame we are not on python3, where you could use the nonlocal keyword to indicate you wanted to operate on the closure variable name. Just as a note in case you prefer this alternative syntax, you can do nonlocal attribute assignment so dectorate.name = name. It really does not make any difference, just a way kinda keep the same name.

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+        if name1 is None:
```
natelust: for more readability and self documenting concider changing the if statements to
if callable(func):
    name1 = func.__name__
elif isintance(func, classmethod) or isinstance(func, staticmethod):
    name1 = func.__func__.__name__
elif isinstance(func, property):
    name1 = func.fget.__name__
else:
    raise ValueError(....)

python/lsst/utils/wrappers.py
TallJimbo: Just for my own education, are you saying I could use `decorate.name = name` to define a local variable within the body of `decorate`, even in Python 2?  I'm not sure I like that better in this case, but it would be something I didn't know was possible.

python/lsst/utils/wrappers.py
TallJimbo: Will `callable(<classmethod instance>)` actually return False?  It's still a callable.

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+
+    C++ template classes are most easily wrapped with a separate Python class
+    for each template type, which results in an unnatural Python interface.
+    TemplateMeta provides a thin layer that connects these Python classes by
+    giving them a common base class and acting as a factory to construct them
+    in a consistent way.
+
+    To use, simply create a new class with the name of the template class, and
+    use ``TemplateMeta`` as its metaclass, and then call ``register`` on each
+    of its subclasses.  Additional type keys for existing classes can be added
```
natelust: You have not mentioned the concept of typekeys yet in this documentation, so the use of the word additional may be confusing to users. Please clarify this concept, perhaps by expanding on what register does and expects. Also, you mention these additional type keys can be created for existing classes. Does this mean they already have to be registered with the class, or simply be declared in python. This is not clear from your wording or example.

python/lsst/utils/wrappers.py
TallJimbo: And I didn't use `isinstance` because I wanted to leave some room for custom descriptors, though admittedly they'd have to work a lot like the builtins to have a chance.

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+    the type keys::
+
+        img = Image(52, 64, dtype=np.float32)
+
+    This simply forwards additional positional and keyword arguments to the
+    wrapped template class's constructor.
+
+    The choice of "dtype" as the name of the template parameter is also
+    configurable, and in fact multiple template parameters are also supported,
+    by setting a ``TEMPLATE_PARAMS`` class attribute to a tuple containing the
```
natelust: Please add the text " class attribute on the abc" to clarify where the user should add this attribute. I.e. having an abc, instance class, and meta class makes the word class a bit overloaded.

python/lsst/utils/wrappers.py
natelust: I agree with Pim's concerns, but also Jim's comment. I wanted to comment to make sure that this does not get lost, and that work is planned for this. I would support some sort of factory method or somethings, just to make it feel more pythonic.

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+        if subclass in self._registry:
+            return True
+        for v in self._registry.values():
+            if issubclass(subclass, v):
+                return True
+        return False
+
+    def __instancecheck__(self, instance):
+        # Special method hook for the isinstance built-in: we return
+        # true for an instance any registered type or true subclass thereof.
```
natelust: The wording does not make sense in this line

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+        attrs["_inherited"] = {k: v for k, v in attrs.items()
+                               if isAttributeSafeToTransfer(k, v)}
+        # The special "TEMPLATE_PARAMS" class attribute, if defined, contains
+        # names of the template parameters, which we use to set those
+        # attributes on registered subclasses and intercept arguments to the
+        # constructor.  This line removes it from the dict of things that
+        # should be inherited while setting a default of 'dtype' if it's not
+        # defined.
+        attrs["TEMPLATE_PARAMS"] = \
+            attrs["_inherited"].pop("TEMPLATE_PARAMS", ("dtype",))
```
natelust: you should check to make sure someone did not define a zero length tuple in TEMPLATE_PARAMS, and give a useful help message if they did. Otherwise construction will succeed, but__call__ will fail

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+        attrs["TEMPLATE_DEFAULTS"] = \
+            attrs["_inherited"].pop("TEMPLATE_DEFAULTS",
+                                    (None,)*len(attrs["TEMPLATE_PARAMS"]))
```
natelust: you should verify that TEMPLATE_DEFAULTS is the same length as TEMPLATE_PARAMS, otherwise if TEMPLATE_DEFAULTS is shorter some of the TEMPLATE_PARAMS may get skipped in the __call__ method as zip will stop at the shortest iterator.

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+
+        Register may only be called once for a given key or a given subclass.
+        """
+        if subclass in self._registry.values():
+            raise ValueError(
+                "This subclass has already registered with another key; "
+                "use alias() instead."
+            )
+        if self._registry.setdefault(key, subclass) != subclass:
+            if len(self.TEMPLATE_PARAMS):
```
natelust: Did you mean to check if len(self.TEMPLATE_PARAMS) == 1

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+                d = {self.TEMPLATE_PARAMS[0]: key}
+            else:
+                d = {k: v for k, v in zip(self.TEMPLATE_PARAMS, key)}
+            raise KeyError(
+                "Another subclass is already registered with {}".format(d)
+            )
+
+        def setattrSafe(name, value):
+            if hasattr(subclass, name):
```
natelust: if you store the getattr to a local variable below this line, you can then use the local variable in the if statement and in the ValueError, potentially saving one lookup and possibly making the format line short enough to fit on one one line and look better formatted.

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+                if not getattr(subclass, name) == value:
+                    raise ValueError(
+                        ("subclass already has a '{}' attribute with "
```
natelust: why is the string in a () before the .format it is just extra un-needed characters. Also of note, you could separate the declaration of the string from the ValueError and format call for nicer line breaks if you wish

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+                    )
+            else:
+                setattr(subclass, name, value)
+
+        if len(self.TEMPLATE_PARAMS) == 1:
+            setattrSafe(self.TEMPLATE_PARAMS[0], key)
+        elif len(self.TEMPLATE_PARAMS) == len(key):
+            for p, k in zip(self.TEMPLATE_PARAMS, key):
+                setattrSafe(p, k)
+        else:
```
natelust: If there is only one template parameter, the length of the key is never checked to be length one, and the below error message will not trigger. This may lead to a confusing error message for the user

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+            )
+
+        for name, attr in self._inherited.items():
+            setattr(subclass, name, attr)
+
+    def alias(self, key, subclass):
+        """Add an alias that allows an existing subclass to be accessed with a
+        different key.
+        """
+        old = tuple(getattr(subclass, p, None) for p in self.TEMPLATE_PARAMS)
```
natelust: This is a silly point, but consider using a name other than old. Old might convey the idea that something new is to come along and replace it. Maybe something like existingKey? This is not something I require, just something to think about

tests/testWrappers.py
```diff
@@ -0,0 +1,386 @@
...
+    def testAliases(self):
+        self.register()
+        self.alias()
+        self.assertEqual(self.ExampleF.dtype, np.float32)
+        self.assertEqual(self.ExampleD.dtype, np.float64)
+        self.assertIn("F", self.Example)
+        self.assertIn("D", self.Example)
+        self.assertEqual(self.Example["F"], self.ExampleF)
+        self.assertEqual(self.Example["D"], self.ExampleD)
+
```
natelust: just to be pedantic it would be nice to see a comparison to the returns from the registered lookups

tests/testWrappers.py
```diff
@@ -0,0 +1,386 @@
...
+        d = self.ExampleD()
+        self.assertIsInstance(f, self.Example)
+        self.assertIsInstance(d, self.Example)
+        self.assertEqual(set(self.Example.__subclasses__()), set([self.ExampleF, self.ExampleD]))
+
+    def testConstruction(self):
+        self.register()
+        f = self.Example(dtype=np.float32)
+        self.assertIsInstance(f, self.Example)
+        self.assertIsInstance(f, self.ExampleF)
```
natelust: for completeness can you test that f is not self.ExampleD

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+            raise ValueError(
+                "key must have {} elements (one for each of {})".format(
+                    len(self.TEMPLATE_PARAMS), self.TEMPLATE_PARAMS
+                )
+            )
+
+        for name, attr in self._inherited.items():
+            setattr(subclass, name, attr)
+
+    def alias(self, key, subclass):
```
natelust: This function will allow a key to be reassigned to a new subclass, is this the behavior we want? What if someone unknowingly re-aliases a subclass, and that breaks other code. Would it be better that the key not be in use and have a key clearing mechanism if you want to be able to reassign a key? or just a reassign method?

tests/testWrappers.py
```diff
@@ -0,0 +1,386 @@
...
+        with self.assertRaises(TypeError):
+            for x in self.ExampleF:
+                pass
+
+    def testAliasUnregistered(self):
+        with self.assertRaises(ValueError):
+            self.Example.alias("F", self.ExampleF)
+        self.assertEqual(len(self.Example), 0)
+        with self.assertRaises(ValueError):
+            self.ExampleF.dtype = "D"
```
natelust: I'm not sure what the point of setting this value is, wont this fail at like 324 just like the assertRaises above?

tests/testWrappers.py
```diff
@@ -0,0 +1,386 @@
...
+        self.assertEqual(len(self.Example), 2)
+        self.assertEqual(set(iter(self.Example)), set([np.float32, np.float64]))
+        self.assertEqual(self.Example.get(np.float64), self.ExampleD)
+        self.assertEqual(self.Example.get(np.int32, False), False)
+
+    def testNoInheritedDictBehavior(self):
+        self.register()
+        f = self.ExampleF()
+        with self.assertRaises(TypeError):
+            len(f)
```
natelust: on python2 for me this line fails the unit test because it raises AttributeError: ExampleF instance has no attribute '__len__' and not a TypeError

tests/testWrappers.py
```diff
@@ -0,0 +1,386 @@
...
+
+    @property
+    def property1(self):
+        return False
+
+
+class DecoratorsTestCase(lsst.utils.tests.TestCase):
+
+    def setUp(self):
+        @lsst.utils.continueClass
```
natelust: you should explicitly import lsst.utils

python/lsst/utils/wrappers.py
```diff
@@ -0,0 +1,363 @@
...
+        # __call__ is invoked when someone tries to construct an instance of
+        # the abstract base class.
+        # If the ABC defines a "TEMPLATE_PARAMS" attribute, we use those strings
+        # as the kwargs we should intercept to find the right type.
+        key = tuple(kwds.pop(p, d) for p, d in zip(self.TEMPLATE_PARAMS,
+                                                   self.TEMPLATE_DEFAULTS))
+        if len(key) == 1:
+            # indices are only tuples if there are multiple elements
+            key = key[0]
+        cls = self._registry.get(key, None)
```
natelust: Ok, this is a subtle behavior, but I will to my best to describe the issue. None is hashable, and can be used as a key in a dictionary. As this code is written it is possible to register (or alias) None to a subclass type, i.e. 
```
Example.register(None, ExampleN)
```
If this is done, then it would be possible in code to do
```
inst = Example()
```
If the TEMPLATE_DEFAULTS has a None state, then the default tuple created would be None. Line 234 would then look up the type registered or aliased to None and construct that. This may be useful as a default constructor option, but in the code snippet above inst would be of the registered type and not what it would appear to be, an instance of the abc. This comment is just to bring up the discussion if we want to allow None such to set a default type, or if we want to filter out None from being a valid key in register and alias.

natelust:
```
In [1]: from future.utils import with_metaclass

In [2]: import lsst.utils

In [3]: class BaseFoo(with_metaclass(lsst.utils.TemplateMeta, object)):
   ...:     def method1(self):
   ...:         return self
   ...:     

In [4]: class FooF:
   ...:     pass
   ...: 

In [5]: import numpy as np

In [6]: BaseFoo.register(np.float32, FooF)

In [7]: BaseFoo(dtype=np.float32)
Out[7]: <__main__.FooF instance at 0x7f24423b8ea8>

In [8]: class FooN:
   ...:     pass
   ...: 

In [9]: BaseFoo.register(None, FooN)

In [10]: BaseFoo()
Out[10]: <__main__.FooN instance at 0x7f2442400830>
```

natelust: If this was intended behavior, well then ignore the discussion

tests/testWrappers.py
```diff
@@ -0,0 +1,386 @@
...
+        self.assertIsInstance(d, self.Example)
+        self.assertEqual(set(self.Example.__subclasses__()), set([self.ExampleF, self.ExampleD]))
+
+    def testConstruction(self):
+        self.register()
+        f = self.Example(dtype=np.float32)
+        self.assertIsInstance(f, self.Example)
+        self.assertIsInstance(f, self.ExampleF)
+        with self.assertRaises(TypeError):
+            self.Example()
```
natelust: As noted above this will not always fail, if a type has been registered with None, and defaults are set to None for TEMPLATE_PARAMS, then this will construct whatever is registered to None.

python/lsst/utils/wrappers.py
TallJimbo: Good catch.  I'm inclined to check for `None` in `register` and disallow it there, but allow it in `alias`, as a way to effectively set a default `dtype` (something @pschella asked about previously).  I'll also add another method to explicitly set the default that will delegate to calling `alias` with `None`, as making that the preferred interface for that would be confusing.

natelust: Also to note, this could happen anytime the defaults are set to something that is already registered with the abc. if BaseFoo.register((1,2,3), ExampleFoo) and TEMPLATE_DEFAULTS were 1,2,3, then BaseFoo() would resolve to ExampleFoo, however I think this maybe a more obvious thing.

python/lsst/utils/wrappers.py
TallJimbo: The parentheses were to make the multi-line string concatenation more explicit, but I've taken your suggestion and moved the declaration of the string to a separate line anyway.

python/lsst/utils/wrappers.py
TallJimbo: If there is only one template parameter, the key should not be a length-one tuple; it should just be whatever object is a key.  Because we don't know what kind of object it is (and it could be something like a `str` that defines `__len__`), I think any attempt to improve that error message could backfire.

python/lsst/utils/wrappers.py
TallJimbo: I've disabled aliasing with an existing key.

tests/testWrappers.py
TallJimbo: Switched to `Exception`, with a comment explaining why.

tests/testWrappers.py
TallJimbo: It was a relic of previous implementation.  I've removed this test.

tests/testWrappers.py
TallJimbo: Fixed to check for `Exception` instead, with a comment explaining why.

python/lsst/utils/wrappers.py
TallJimbo: I decided to disallow `None` in both `register` and `alias`. There's already a way to set defaults, and we don't need another.

python/lsst/utils/wrappers.py
natelust: Well in the body of the function you would have to refer to decorate.name to get at the variable. It is just attaching something to the instance of the function

python/lsst/utils/wrappers.py
natelust: interestingly, a classmethod that is already attached to a class will return true in callable, a free class/static method function that is to be monkey patched in later will not.

TallJimbo: Ah, that's because trying to get a `classmethod` that's attached to a class with `getattr` will invoke magic that returns a bound method object instead of the `classmethod` itself (`classmethod` is a rare descriptor that also does magic when you access it on a class).  This is also why I had use `__dict__` to get some attributes in `continueClass`.

PR #30, opened 2017-03-07
DM-9715: cppIndex should raise Python's built-in IndexError

r-owen:
In order to test this easily I also added a pybind11 wrapper.
for the function and changed the test from C++ to Python (in testPybind11.py)


include/lsst/utils/python.h
```diff
@@ -24,6 +24,7 @@
 #ifndef LSST_UTILS_PYTHON_H
 #define LSST_UTILS_PYTHON_H
 
+#include "Python.h"  // for PyErr_...
```
TallJimbo: pybind11.h is guaranteed to import this; we don't need to do it explicitly.

TallJimbo:
I don't see `python/lsst/utils/python.cc`. Did you forget to add it?

Also, I wonder if we could now move the implementations of `cppIndex` to that file instead of making them `inline`.

include/lsst/utils/python.h
```diff
@@ -82,7 +86,8 @@ inline std::size_t cppIndex(std::ptrdiff_t size, std::ptrdiff_t i) {
     if (i < 0 || i >= size) {
         std::ostringstream os;
         os << "Index " << i_orig << " not in range [" << -size << ", " << size - 1 << "]";
-        throw LSST_EXCEPT(lsst::pex::exceptions::OutOfRangeError, os.str());
+        PyErr_SetString(PyExc_IndexError, os.str().c_str());
```
pschella: Why not just throw `pybind11::index_error`?

include/lsst/utils/python.h
```diff
@@ -108,7 +113,8 @@ inline std::pair<std::size_t, std::size_t> cppIndex(std::ptrdiff_t size_i, std::
         os << "Index (" << i << ", " << j << ") not in range ["
            << -size_i << ", " << size_i - 1 << "], ["
            << -size_j << ", " << size_j - 1 << "]";
-        throw LSST_EXCEPT(lsst::pex::exceptions::OutOfRangeError, os.str());
+        PyErr_SetString(PyExc_IndexError, os.str().c_str());
```
pschella: Same here. Throw `pybind11::index_error`.

tests/testPybind11.py
```diff
@@ -124,6 +125,72 @@ def testOverloads(self):
         self.assertEqual(_example.getName(int(1)), "int")
         self.assertEqual(_example.getName(float(1)), "double")
 
+    def testCppIndex1Axis(self):
```
pschella: Wrapping / testing this in Python looks a bit weird. But ok.

include/lsst/utils/python.h
r-owen: I prefer explicit, but if you insist

r-owen:
I initially did forget to push python.cc but it's there now (my apologies; I spotted the error when Jenkins failed)

r-owen:
@TallJimbo do you *really* want me to move the implementation of `cppIndex` to `python/utils/python.cc`? That seems a very odd place for it. Putting it in `src/python.cc` would be more obvious, I think, but heavyweight for such simple function. inline seems at least semi-logical for something so light and that will appear in loops.

tests/testPybind11.py
r-owen: I agree. Maybe it would be better to not hoist the symbol into the lsst.utils namespace?

include/lsst/utils/python.h
r-owen: Great idea. That's simpler, saves any impulse to import Python.h and allows me to use `os.str()` instead of `os.str().c_str()`.

TallJimbo:
Putting the implementation in `src` would not work - that would require `libutils.so` to link against Python.  It seemed a bit long for an `inline`, but not so long as to be unreasonable.  It's fine to leave it as-is as well.

PR #33, opened 2017-05-02
DM-5637: Enable travis flake8 and make compliant

timj:
[DM-5637] Add .travis.yml and setup.cfg to run flake8.


timj:
DM-5637 requested flake8 tests on branches before we will let them merge. If we accept this PR these tests will be enabled for the utils package. Thanks to @jmatt for setting this up. I have made `utils` compliant (with the current rules, not the 80 character rule).

timj:
@jmatt apologizes for pushing the button prematurely. Feel free to review it and we can revert if this messes up @parejkoj with #32.

jdswinbank:
@timj suggests that I should comment here.

The implementation looks good, and this is a great feature. I'd rather that it was properly documented (in the Dev Guide, ie [here](https://developer.lsst.io/build-ci/ci_overview.html)) before it goes live on active repos, though: I worry that our CI infrastructure is already pretty confusing, and adding more features without docs isn't going to help.

jdswinbank:
(I realise it went live on utils by mistake, and I'm not worried about that — just like to avoid a wider roll-out.)

PR #34, opened 2017-05-03
ignore astropy.log when testing file descriptor leaks

danielsf:
Some users of LSST Simulations packages have been seeing testFileDescriptorLeaks() failures because they have configured astropy to log error messages, warnings etc. to an astropy.log file, which is still open at the point where testFileDescriptorLeaks() runs.  This PR adds astropy.log to the set of files ignored by testFileDescriptorLeaks().

### DIFF
--- a/python/lsst/utils/tests.py
+++ b/python/lsst/utils/tests.py
@@ -170,7 +170,8 @@ class MemoryTestCase(unittest.TestCase):
         # Some files are opened out of the control of the stack.
         now_open = set(f for f in now_open if not f.endswith(".car") and
                        not f.endswith(".ttf") and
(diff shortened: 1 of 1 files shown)


danielsf:
@timj So... do I need to flake8 the whole file, or just my line (and where to I find the explanation of what failed the test)?

timj:
it was fully compliant before your patch so it was your patch that caused the flake8 warning. For now you can see the output of the flake8 by clicking on the travis job "Details" button. It tells you you have a problem with your operator line breaking. I don't know if it's possible in theory for Travis to send a message to a PR with details of the failure. @jmatt ?

timj:
Please squash the two commits before merging.

PR #35, opened 2017-05-03
DM-10448: Rewrap to 79 characters

timj:
This commit implements RFC-107 for the utils package. The good news is that autopep8 does work. The bad news is that:

1. it sometimes makes some very bad aesthetic choices. Especially if multiple parentheses are involved in the block.
2. it doesn't completely fix everything so there will be lines left broken. In particular doc strings (or any quoted strings) will not be fixed up.

Therefore these commits are about 30% automated and the rest manual.

### DIFF
--- a/python/lsst/utils/__init__.py
+++ b/python/lsst/utils/__init__.py
@@ -1,4 +1,4 @@
-# 
+#
 # LSST Data Management System
 #
 # Copyright 2008-2017  AURA/LSST.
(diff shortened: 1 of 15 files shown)


parejkoj:
I think we can probably close this without merging?

timj:
Yes. The docstrings and comments will need to be shortened but we'll have to redo those later. We really want PyCQA/pycodestyle#674 to be merged.

timj:
Abandon this. We will stick with 110 for code and 79 for docstrings now that pycodestyle has accepted our patch.

PR #36, opened 2017-06-08
Add automatic backtrace on segfault.

kfindeisen:


include/lsst/utils/Backtrace.h
```diff
@@ -0,0 +1,195 @@
...
+private:
+    Backtrace() {
+        // Register abort handlers
+        signal(SIGABRT, signalHandler);
+        signal(SIGSEGV, signalHandler);
+        signal(SIGILL, signalHandler);
+        signal(SIGFPE, signalHandler);
+    }
+
+    static void signalHandler(int signum) noexcept {
```
kfindeisen: Also, given that everywhere else you try to stick to C (`.h` headers, `malloc`, `char*`), is it ok to have `noexcept` here? (Or is it specifically to make the code more C-like by removing exception support?)

src/Backtrace.cc
```diff
@@ -0,0 +1,36 @@
...
+ * see <https://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#include "lsst/utils/Backtrace.h"
+
+namespace lsst {
+namespace utils {
+
+namespace {
+Backtrace& backtrace = Backtrace::get();
```
kfindeisen: I assume this is to force instantiation of the singleton, but does it actually work? I could see a compiler optimizing this out as an unused variable.

tests/backtrace.cc
```diff
@@ -0,0 +1,39 @@
...
+#include "pybind11/pybind11.h"
+
+namespace py = pybind11;
+
+void generateSegfault() {
+    raise(SIGSEGV);
+}
+
+PYBIND11_PLUGIN(_backtrace) {
+    py::module mod("_backtrace");
```
kfindeisen: Why use this roundabout method instead of just testing in C++?

include/lsst/utils/Backtrace.h
```diff
@@ -0,0 +1,195 @@
...
+    static constexpr unsigned int FUNCTION_NAME_SIZE = 1024;
+
+    // No copying and moving
+    Backtrace(Backtrace const&) = delete;
+    Backtrace(Backtrace&&) = delete;
+    Backtrace& operator=(Backtrace const&) = delete;
+    Backtrace& operator=(Backtrace&&) = delete;
+
+    /// Get a reference to the singleton
+    static Backtrace& get() {
```
kfindeisen: Can be declared `noexcept`, but see question for `signalHandler`.

include/lsst/utils/Backtrace.h
```diff
@@ -0,0 +1,195 @@
...
+        static Backtrace st;
+        return st;
+    }
+
+private:
+    Backtrace() {
```
kfindeisen: Can be declared `noexcept` (like all C functions, `signal` is non-throwing).

kfindeisen:
A couple more thoughts:
- Does `Backtrace` need a destructor that unregisters the signal handler?
- Should `Backtrace` be `final`? Like with `noexcept`, I'm not sure whether this would make the code more or less C-like...

include/lsst/utils/Backtrace.h
```diff
@@ -0,0 +1,195 @@
...
+
+#ifndef LSST_UTILS_BACKTRACE_H
+#define LSST_UTILS_BACKTRACE_H
+
+#include <stdio.h>
+#include <signal.h>
+#include <execinfo.h>
+#include <errno.h>
+#include <cxxabi.h>
+#include <stdlib.h>
```
r-owen: I agree with @kfindeisen. If you are sure that all linux and Mac C++ compilers will have these headers, then this seems a reasonable approach (though a code comment explaining why you are sure would be in order). If not, then you may need to figure out some kind of test and provide a null implementation (perhaps with a warning printed to stderr) if the conditions are not met. Or give up on the idea.

include/lsst/utils/Backtrace.h
```diff
@@ -0,0 +1,195 @@
...
+
+        size_t funcNameSize = FUNCTION_NAME_SIZE;
+        char* funcName = (char*)malloc(funcNameSize * sizeof(char));
+
+        for (unsigned int i = 0; i < addressLength; i++) {
+// demangle names (if we can)
+#ifdef __APPLE__
+            // OSX style stack trace
+            char* beginName = nullptr;
+            char* beginOffset = nullptr;
```
r-owen: Why is the code so C-like? Are you worried about running out of memory if you use std::string?

include/lsst/utils/Backtrace.h
```diff
@@ -0,0 +1,195 @@
...
+/**
+ *  Singleton, enables automatic backtraces on the following signals:
+ *
+ *  - SIGABRT
+ *  - SIGSEGV
+ *  - SIGILL
+ *  - SIGFPE
+ *
+ *  @note Uses low level malloc and fprintf since higher level constructs
+ *  may not be available when a signal is received.
```
kfindeisen: @r-owen see here.

include/lsst/utils/Backtrace.h
```diff
@@ -0,0 +1,195 @@
...
+                }
+
+                // mangled name is now in [beginName, beginOffset)
+                // and caller offset in [beginOffset, endOffset).
+                int status = 0;
+                funcName = abi::__cxa_demangle(beginName, funcName, &funcNameSize, &status);
+                char* fname = beginName;
+                if (status == 0) {
+                    fname = funcName;
+                }
```
r-owen: I find it confusing to have `fname` and `funcName` as different variables

include/lsst/utils/Backtrace.h
```diff
@@ -0,0 +1,195 @@
...
+
+        for (unsigned int i = 0; i < addressLength; i++) {
+// demangle names (if we can)
+#ifdef __APPLE__
+            // OSX style stack trace
+            char* beginName = nullptr;
+            char* beginOffset = nullptr;
+            for (char* p = symbolList[i]; *p; ++p) {
+                if ((*p == '_') && (*(p - 1) == ' ')) {
+                    beginName = p - 1;
```
r-owen: It seems unsafe to test `*(p - 1)` without first insuring that p has been incremented past its starting position.

Why is `beginName` set to `p - 1` if this condition is met? It seems like it will assuredly be set to a space. A code comment explaining what is going on would be appreciated. The code below is equally mysterious to me.

include/lsst/utils/Backtrace.h
```diff
@@ -0,0 +1,195 @@
...
+                } else if (*p == '+') {
+                    beginOffset = p - 1;
+                }
+            }
+
+            if (beginName && beginOffset && (beginName < beginOffset)) {
+                *beginName++ = '\0';
+                *beginOffset++ = '\0';
```
r-owen: Are you sure this is guaranteed to do what you expect? The following gives me doubts: https://stackoverflow.com/questions/4445706/post-increment-and-pre-increment-concept

RobertLuptonTheGood:
I also am worried about portability. The safe thing is to add a test in sconsUtils, 

        R

lib/SConscript
```diff
@@ -1,3 +1,16 @@
...
-from lsst.sconsUtils import scripts
-scripts.BasicSConscript.lib()
+from lsst.sconsUtils import scripts, env
+
+# Backtrace depends on system headers which may not be
+# available on all platform / compiler combinations.
+# Check if headers exist and if not don't build it.
+noBuildList = []
+conf = Configure(env)
+for header in ['cxxabi.h', 'execinfo.h', 'signal.h']:
```
kfindeisen: `signal.h` should always be available: http://en.cppreference.com/w/cpp/header#C_compatibility_headers

include/lsst/utils/Backtrace.h
pschella: Yes, that is exactly the reason. Although nothing is probably safe if we run out of memory we should at least make some attempt to prevent it. Ideally we would fall back to `backtrace_symbols_fd` but that doesn't seem to be universally available. We could probably use `std::string` if we are careful, but I don't think it adds much here and I definitely want to avoid `iostream`.

tests/backtrace.cc
pschella: No real reason, I'm still a bit undecided on how best to test this. Given that it requires segfaulting which will kill the test. Perhaps run it in a subprocess? Might not be worth it though. Suggestions welcome.

include/lsst/utils/Backtrace.h
```diff
@@ -0,0 +1,72 @@
...
+#ifndef LSST_UTILS_BACKTRACE_H
+#define LSST_UTILS_BACKTRACE_H
+
+#include <cerrno>
+#include <cstdio>
+#include <cstdlib>
+#include <csignal>
+
+#include <cxxabi.h>
+#include <execinfo.h>
```
r-owen: I suggest adding a comment above these two includes explaining that Backtrace will not be built unless these files are found, and explaining that the test is in `lib/SConscript`.

I assume even if the library is not built, that `Backtrace.h` will be visible in the include path, so users can shoot themselves in the foot by including this file -- resulting in code that will mysteriously fail to compile on some systems. Is there a way to hide this header file, so external users of `utils` cannot find it, e.g. by putting it in `src` instead of `include`? I realize that makes it more difficult to build the pybind11 wrapper.

src/Backtrace.cc
```diff
@@ -0,0 +1,284 @@
...
+    *end = NULL;
+    char *lastSpace = NULL;
+    while (*input != '\0') {
+        if (*input == ' ') {
+            lastSpace = input;
+        } else if (*input == '_' && lastSpace == input - 1) {
+            *start = input;
+        } else if (*input == '+' && lastSpace == input - 1) {
+            *end = input - 1;
+        }
```
r-owen: The two cases above both refer to `input - 1` without insuring they exist first. Simply testing that lastSpace is not NULL before examining `input - 1` would solve that, making the code more robust, or at least more obviously safe.

src/Backtrace.cc
```diff
@@ -0,0 +1,284 @@
...
+
+#else  // Unknown stack trace style (no demangling)
+
+char *demangleAndPrint(char *input, char *buffer, size_t *bufferSize) noexcept {
+    fprintf(stderr, "  %s\n", input);
+    return buffer;
+}
+
+#endif  // __linux__ && __GNUC__
+#endif  // __APPLE__ && __clang__
```
r-owen: I'm curious how bad the mangled stack trace really is. if it's just a case of some C++ symbols that the user could demangle manually using `c++filt` I'd be inclined to skip all the above and just use the simple code. However, I'm guessing it's worse.

I worry because this is a lot of very complicated code to handle a very rare situation where any output is better than none. I also worry because, despite the large amount of code, this still doesn't handle the case of linux with clang, which I hope be one of our standard build platforms.

tests/backtrace.cc
r-owen: I have no idea how to test this. It is one of my big worries, especially given the complicated code to format the stack trace -- is there any to test that formatting?

I'm inclined to agree with @kfindeisen that if you can nearly as easily test in C++ it would be better to do so and get rid of this wrapper.

include/lsst/utils/Backtrace.h
pschella: This sounds reasonable, but I'm not sure if there is precedent for adding header files in the `src` directory?

src/Backtrace.cc
pschella: Fair point, although in that case the return value would always be `false` anyway (because `*end` would be `NULL`), I agree it is probably more obviously safe.

src/Backtrace.cc
pschella: Yes, indeed it is only the C++ symbols, and they can be run through `c++filt`. This may be acceptable, however remember that one does need to run `c++filt` on the same system. This may be a bit annoying when getting traces from a crashed cluster run or from an end user. But that may be a price worth paying for decreased complexity.

src/Backtrace.cc
r-owen: Interesting. I was worried `input - 1` might point to garbage. I realize there are probably reasons why it would not, but are those reasons obvious?

r-owen:
If running a program that segfaults in a subprocess will allow you to verify the formatting, then perhaps you could code a trivial C++ program to segfault and spawn that from a Python unit test?

include/lsst/utils/Backtrace.h
r-owen: I agree putting a header file in `src` seems undesirable. Perhaps one of the following would do the trick:
- Move the header file to a `detail` subdirectory. Then it's clearly not part of the public API. Also move the pybind11 wrapper to `utils.detail`, if you keep the wrapper, so generateSegfault is not public.
- If you plan to get rid of the pybind11 wrapper then eliminate the header file and add any needed includes to the src file.

src/Backtrace.cc
pschella: It may very well point to garbage, but the only way it would pass the test is if it would be `NULL` (because for `input - 1` to be invalid the first character would have to be `+` at which point `lastSpace == NULL` and then `*end == NULL` and thus the result is `false`. This of course all assumes the input is a valid `\0` terminated string.

r-owen: The issue I have with this is it assumes too much about the contents of the data (that for `input - 1` to be invalid the first character must be `+`). I realize you have to assume things about the format to write this code in the first place, but by coding defensively the worst that can happen if the format changes is for the output to be weird -- you will never read garbage.

src/Backtrace.cc
```diff
@@ -0,0 +1,284 @@
...
+            case ')':
+                closingBrace = input;
+                break;
+            case '[':
+                *startAddress = input + 1;
+                break;
+            case ']':
+                closingBracket = input;
+                break;
+        }
```
kfindeisen: Is there any way you might get, e.g., nested parentheses?

src/Backtrace.cc
```diff
@@ -0,0 +1,284 @@
...
+ * Substrings are marked with '\0' when found.
+ *
+ * @param input entry produced by backtrace_symbol
+ * @param startName start of the mangled name (substring terminated by `\0')
+ * @param startOffset start of the offset (substring terminated by `\0')
+ * @param startAddress start of the address (substring terminated by `\0')
+ *
+ * @return true if all substrings are found.
+ */
+bool findMangledName(char *input, char **startName, char **startOffset, char **startAddress) noexcept {
```
kfindeisen: Why does this `findDemangledName` need a different parameter list from the first one? Couldn't you make them the same, and then factor out an OS-independent `demangleAndPrint`?

src/Backtrace.cc
```diff
@@ -0,0 +1,284 @@
...
+            case ']':
+                closingBracket = input;
+                break;
+        }
+        ++input;
+    }
+
+    if (*startName && *startOffset && *startAddress && closingBrace && closingBracket) {
+        // All intervals are found, now split them into substrings
+        *(*startName - 1) = '\0';
```
kfindeisen: Like with the discussion above, it might be worth noting that `startName-1 ≥ input` (and analogously for the other `start` pointers) regardless of the string format.

src/Backtrace.cc
```diff
@@ -0,0 +1,284 @@
...
+ * @param input entry produced by backtrace_symbol
+ * @param buffer allocated with malloc for output of the demangled name
+ *               if the name does not fit the bufferSize is increased with realloc
+ *               the user is responsible for freeing the buffer
+ * @param bufferSize size of the name buffer (increased if needed)
+ *
+ * @return buffer (possibly reallocated) buffer for demangled name
+ */
+char *demangleAndPrint(char *input, char *buffer, size_t *bufferSize) noexcept {
+    int status;
```
kfindeisen: Move this to the scope where it's used.

src/Backtrace.cc
```diff
@@ -0,0 +1,284 @@
...
+    // allocate storage for demangled name
+    // this will be reallocated by abi::__cxa_demangle in demangleAndPrint
+    // if storage is insufficient, but if it is sufficient
+    // this reduces the number of allocations
+    size_t nameSize = NAME_SIZE_ESTIMATE;
+    char *name = (char *)malloc(nameSize * sizeof(char));
+    if (name == NULL) {
+        fprintf(stderr, "[malloc] cannot dump trace, probably out of memory\n");
+        raiseWithDefaultHandler(signum);
+    }
```
kfindeisen: What happens if `name` is allocated successfully, but then `abi::__cxa_demangle` tries and fails to reallocate it? Do you keep the old buffer, just with truncated contents?

src/Backtrace.cc
r-owen: I disagree that the second and third branch require `lastSpace != NULL`. Consider the following trivial example:  If the first character of `input` is "_" then the second branch will be followed on that first character (despite `lastSpace being NULL) and the test will reference the undefined value `input - 1`.

I think the code *must* be changed to explicitly require `lastSpace != NULL` before reading `input - 1`. One way to do this is nest those tests in a new block: `} else if (lastSpace != NULL) {...second and third tests...}`

r-owen: But in any case @kfindeisen raises an excellent point about buffer overflow.

kfindeisen: @r-owen In your example, how is the condition `lastSpace == input - 1` satisfied? I assume `1` is not a valid address?

r-owen: If `input - 1` is garbage that garbage may well be NULL. I really don't see why we are arguing about this. Proper behavior can easily be assured by by explicitly testing that `lastSpace` is not NULL -- thus assuring that a space has been seen. It makes the code more robust against changes in formatting. Otherwise you may or may not fall through this test, giving inconsistent output from one run to the next (or on one system that reliably has NULL at input - 1 vs. another that does not).

pschella: Please note that nowhere was I arguing against adding the safety check, and I will put it in. I was just arguing that I think the code is in fact safe without it (assuming **only** that the input is a valid `\0` terminated string, which I think is a valid assumption). But I will include it nonetheless for clarity.

src/Backtrace.cc
pschella: I think you do, but the manual isn't clear about that.

tests/testExecutables.py
```diff
@@ -6,7 +6,7 @@ class UtilsBinaryTester(lsst.utils.tests.ExecutablesTestCase):
     pass
 
 
-EXECUTABLES = None
+EXECUTABLES = ["testexe.sh"]
```
pschella: `testexe.sh` is the only executable in the test directory. I didn't want to have to add yet another thing to `sconsUtils` to tell it to ignore this test, so instead I provide the list here.

pschella: But it is no longer needed, now that the test structure has changed. So I'm putting `None` back.

PR #37, opened 2017-06-22
DM-11009: Automatic backtrace printing is unhelpful

pschella:


RobertLuptonTheGood:
Is using `boost::regex` absolutely necessary?  We're trying to reduce our dependence on boost (ideally to near zero)

jdswinbank:
I asked this on Slack:

> swinbank [08:07]
> Why the switch to Boost? I thought we were trying to avoid that where possible. I mean, the answer is presumably some variant of "because it doesn't crash", but we should at least explain that in the commit message I think.
> 
> pschella [08:40] 
> Ok, I’ll reword the commit message. I think the problem is that the compiler version we have on lsst-dev doesn’t fully support the `std::regex` library (that is why we haven’t yet converted our other code to use it instead of `boost::regex`). So we will have to live with this until we finally switch to `devtoolset-6`.

timj:
I'm happy with boost::regex until we can finally upgrade our compiler.

PR #38, opened 2017-07-12
tickets/DM-7477

natelust:
If a type ABC is declared with a default key, copy over static and
class methods from the class registered to the default key to the
type ABC.

### DIFF
--- a/python/lsst/utils/wrappers.py
+++ b/python/lsst/utils/wrappers.py
@@ -23,6 +23,7 @@
 from __future__ import absolute_import, division, print_function
 
 import sys
+import types
 
 __all__ = ("continueClass", "inClass", "TemplateMeta")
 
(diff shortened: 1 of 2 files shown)


python/lsst/utils/wrappers.py
```diff
@@ -322,6 +323,34 @@ def register(self, key, subclass):
...
+        # make the static methods available through the ABC
+        if self.TEMPLATE_DEFAULTS:
+            match = False
+            if len(self.TEMPLATE_DEFAULTS) == 1:
+                if key == self.TEMPLATE_DEFAULTS[0]:
+                    match = True
+            elif key == self.TEMPLATE_DEFAULTS:
+                match = True
+
+            if match:
```
PaulPrice: Can simplify the above logic:

        if self.TEMPLATE_DEFAULTS:
            defaults = (self.TEMPLATE_DEFAULTS[0] if len(self.TEMPLATE_DEFAULTS) == 1 else
                        self.TEMPLATE_DEFAULTS)
            if key == defaults:

python/lsst/utils/wrappers.py
```diff
@@ -322,6 +323,34 @@ def register(self, key, subclass):
...
+                # In the following if statements, the explicit lookup in
+                # __dict__ must be done, as a call to getattr, returns the
```
PaulPrice: Need to drop the second comma on line 338:

    # __dict__ must be done, as a call to getattr returns the

python/lsst/utils/wrappers.py
```diff
@@ -322,6 +323,34 @@ def register(self, key, subclass):
...
+                # method. The static methods must be transfered to the ABC
+                # in this unbound state, so that python will still see them
+                # as static methods and not attempt to pass self. The class
+                # methods must be transfered to the ABC as a bound method
+                # so that the correct cls be called with the class method
+                for obj in subclass.__dict__:
+                    # copy over the static methods
+                    if isinstance(subclass.__dict__[obj],
+                                  types.BuiltinFunctionType) or\
+                       isinstance(subclass.__dict__[obj], staticmethod):
```
PaulPrice: I suggest:
                    
                for objName in subclass.__dict__:
                    obj = subclass.__dict__[objName]
                    # copy over the static methods
                    if isinstance(obj, (types.BuiltinFunctionType, staticmethod)):

etc. This saves multiple lookups and removes that (evil) backslash.

python/lsst/utils/wrappers.py
```diff
@@ -322,6 +323,34 @@ def register(self, key, subclass):
...
+                        setattr(self, obj, subclass.__dict__[obj])
```
PaulPrice: Add a `continue` here, since the `if` below will never fire?

PR #43, opened 2017-11-18
Add /proc to ignore list for open file descriptors

kennywlo:


python/lsst/utils/tests.py
```diff
@@ -169,6 +169,7 @@ def testFileDescriptorLeaks(self):
 
         # Some files are opened out of the control of the stack.
         now_open = set(f for f in now_open if not f.endswith(".car") and
+                       not f.startswith("/proc") and
```
ktlim: Hm.  I'd rather have this be `/proc/`, if you don't mind.

PR #45, opened 2017-12-18
DM-13046: Add random component to getTempFilePath filename

timj:
This fixes getTempFilePath such that there is a little bit of randomness in the constructed name. This has two major benefits:

* meas_base can have two test classes in the same file sharing the same test method names.
* you can now have nested context managers with the same file suffix. This I think will make @danielsf very happy since it was clear that lots of sims tests would have benefited from this.

### DIFF
--- a/tests/test_getTempFilePath.py
+++ b/tests/test_getTempFilePath.py
@@ -27,6 +27,18 @@ import time
 import lsst.utils.tests
 
 
+"""
+This file contains tests for lsst.utils.tests.getTempFilePath.
+
+The TestNameClashN classes are used to check that getTempFilePath
+does not use the same name across different test classes in the same
+file even if they have the same test methods. They are distinct classes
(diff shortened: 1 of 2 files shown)


tests/test_getTempFilePath.py
```diff
@@ -61,6 +62,42 @@ def runLevel3(self, funcName):
         self.runLevel2(funcName)
 
 
+class TestNameClash1(unittest.TestCase):
+    """The TestNameClashN classes are used to check that getTempFilePath
+    does not use the same name across different test classes in the same
+    file even if they have the same test methods."""
```
PaulPrice: Shouldn't this be a module-level docstring or comment, since it's about a group of classes?

Point out you're attempting to trigger a race condition?

timj: Yes. Can do that.

PR #46, opened 2017-12-18
Enable flag to indicate whether temp file is expected to be created

timj:
Sometimes you want the temp file path but that file will not exist when the with block completes. This change adds a new optional argument to allow this. Now if a file exists when it should not, or if a file should exist but does not, there is a RuntimeError.

### DIFF
--- a/python/lsst/utils/tests.py
+++ b/python/lsst/utils/tests.py
@@ -355,16 +355,22 @@ def findFileFromRoot(ifile):
 
 
 @contextmanager
-def getTempFilePath(ext):
+def getTempFilePath(ext, expectOutput=True):
     """!Return a path suitable for a temporary file and try to delete the file on success
 
     If the with block completes successfully then the file is deleted, if possible;
     failure results in a printed warning.
+    If a file is remains when it should not, a RuntimeError exception is raised. This
+    exception is also raised if a file is not present on context manager exit when one
+    is expected to exist.
     If the block exits with an exception the file if left on disk so it can be examined.
     The file name has a random component such that nested context managers can be used
     with the same file suffix.
 
     @param[in] ext  file name extension, e.g. ".fits"
+    @param[in] expectOutput If true, a file should be created within the context manager.
+                            If false, a file should not be present when the context manager
+                            is exited.
     @return path for a temporary file. The path is a combination of the caller's file path
     and the name of the top-level function, as per this simple example:
     @code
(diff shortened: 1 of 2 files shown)


python/lsst/utils/tests.py
```diff
@@ -355,16 +355,22 @@ def findFileFromRoot(ifile):
 
 
 @contextmanager
-def getTempFilePath(ext):
+def getTempFilePath(ext, expectOutput=True):
     """!Return a path suitable for a temporary file and try to delete the file on success
 
     If the with block completes successfully then the file is deleted, if possible;
     failure results in a printed warning.
+    If a file is remains when it should not, a RuntimeError exception is raised. This
```
danielsf: extraneous "is" in the docstring here

timj: Thanks. I'm wondering if this will break Sims so I'm running a Jenkins test. I know that there were tests where you wanted this functionality.

danielsf: I don't think it will break sims....  If it does, let me know and I will fix sims.  This is probably better than whatever hack I have in place to mimic this functionality now.

python/lsst/utils/tests.py
```diff
@@ -405,14 +411,32 @@ def runTest(self):
...
+        except OSError:
+            pass
+
     yield outPath
-    if os.path.isfile(outPath):
+
+    fileExists = os.path.exists(outPath)
+    if expectOutput:
+        if not fileExists:
+            raise RuntimeError("Temp file expected named {} but none found".format(outPath))
```
PaulPrice: Why is this such a big deal? I expected I would create the file myself, but it turns out I didn't.
This is going to be called in the process of handling an exception, and if you're going to throw another exception it will mask the first. Consider:

    with getTempFilePath(".foo", True):
        raise RuntimeError("Some very important message")

With the code as it stands, the above produces a `RuntimeError` from `getTempFilePath`, and the "very important message" is lost.

PaulPrice: Actually, I'm wrong. The original exception won't be masked: the `yield` isn't embedded in a `try` block.

timj:
Regarding the "big deal", I think that in a test environment the warning doesn't help you because no-one sees it. Also, if we know the file isn't going to come the warning is simply annoying. If there are rare cases where the test will run but the file might not turn up then we can pretty easily add a `is None` test for `expectOutput`. For the tests we have I wouldn't expect there to be any effect. We have found one problem in a sims package because there was a typo in a copy and paste.

PR #47, opened 2018-02-01
DM-10066: Provide utility function for wrapping operator<<

kfindeisen:
This PR adds two function templates to `utils/python.h`, one for wrapping the stream output operator as a nullary method that returns a string, and one to wrap `std::hash` as `__hash__` (the latter is technically out of scope for the ticket, but since there's only one class that specializes `std::hash` at present it didn't seem worth it to open another). Both templates are designed to not require explicit template arguments.

### DIFF
--- a/include/lsst/utils/python.h
+++ b/include/lsst/utils/python.h
@@ -88,6 +88,23 @@ void addOutputOp(PyClass &cls, std::string const &method) {
 }
 
 /**
+ * Add `__hash__` method implemented by `std::hash`.
+ *
+ * @tparam PyClass The pybind11 class_ type. The wrapped class must
+ *                 have an enabled specialization of `std::hash`.
+ *
+ * @param cls The `PyClass` object to which to add a wrapper.
+ */
+template <class PyClass>
+void addHash(PyClass &cls) {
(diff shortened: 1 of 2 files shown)


include/lsst/utils/python.h
```diff
@@ -115,6 +115,43 @@ inline std::pair<std::size_t, std::size_t> cppIndex(std::ptrdiff_t size_i, std::
...
+ *
+ * @tparam PyClass The pybind11 class_ type. The wrapped class must
+ *                 support << as a stream output operator.
+ *
+ * @param cls The `PyClass` object to which to add a wrapper.
+ * @param method The name of the method to implement. Should be `"__str__"` or
+ *               `"__repr__"`.
+ */
+template <class PyClass>
+void wrapOutputOp(PyClass &cls, std::string const &method) {
```
pschella: I'd like consistent naming. We also have `addSharedPtrEquality` in the same file, so `add` seems better than `wrap`.

kfindeisen: So just to be clear, you want `addOutputOp` and `addHash`?

PR #51, opened 2018-06-05
DM-11725: Improve code coverage and numpydoc

timj:
I'm testing code coverage testing with pytest and as part of that I ended up improving test coverage in utils. I also tweaked the numpydoc a bit whilst I was there and ensured it is 79 column compliant.

### DIFF
--- a/python/lsst/utils/tests.py
+++ b/python/lsst/utils/tests.py
@@ -63,14 +63,21 @@ open_files = set()
 
 
 def _get_open_files():
-    """Return a set containing the list of open files."""
+    """Return a set containing the list of files currently open in this
+    process.
+
+    Returns
+    -------
+    open_files : `set`
+        Set containing the list of open files.
+    """
     if psutil is None:
         return set()
     return set(p.path for p in psutil.Process().open_files())
 
 
 def init():
-    """Initialize the memory tester"""
+    """Initialize the memory tester and file descriptor leak tester."""
     global memId0
     global open_files
     if dafBase:
@@ -80,7 +87,25 @@ def init():
 
 
 def run(suite, exit=True):
-    """Exit with the status code resulting from running the provided test suite"""
+    """Run a test suite and report the test return status to caller or shell.
+
+    .. note:: Deprecated in 13_0
+              Use `unittest.main()` instead, which automatically detects
+              all tests in a test case and does not require a test suite.
+
+    Parameters
+    ----------
+    suite : `unittest.TestSuite`
+        Test suite to run.
+    exit : `bool`, optional
+        If `True`, Python process will exit with the test exit status.
+
+    Returns
+    -------
+    status : `int`
+        If ``exit`` is `False`, will return 0 if the tests passed, or 1 if
+        the tests failed.
+    """
 
     warnings.warn("lsst.utils.tests.run() is deprecated; please use unittest.main() instead",
                   DeprecationWarning, stacklevel=2)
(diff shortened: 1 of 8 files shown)


python/lsst/utils/tests.py
```diff
@@ -63,14 +63,20 @@
 
 
 def _get_open_files():
-    """Return a set containing the list of open files."""
+    """Return a set containing the list of open files from this process.
```
r-owen: What does "from this process" mean? Files that were opened by this process? Please reword or clarify in a longer description following the brief description.

python/lsst/utils/tests.py
```diff
@@ -63,14 +63,20 @@
...
+        Set containing the list of open files.
+    """
     if psutil is None:
         return set()
     return set(p.path for p in psutil.Process().open_files())
 
 
 def init():
-    """Initialize the memory tester"""
+    """Initialize the memory tester and leak tester."""
```
r-owen: What is the difference between the memory tester and the leak tester? I thought the memory tester tested for memory leaks. Please reword and/or clarify in a longer description.

python/lsst/utils/tests.py
```diff
@@ -80,7 +86,28 @@ def init():
 
 
 def run(suite, exit=True):
-    """Exit with the status code resulting from running the provided test suite"""
+    """Exit with the status code resulting from running the provided test
+    suite
```
r-owen: This is a pre-existing issue, but the returned value is apparently the test exit status only if `exit` is True. Please consider changing this to something like `Run a test suite`

python/lsst/utils/tests.py
```diff
@@ -80,7 +86,28 @@ def init():
...
+
+    .. note:: Deprecated in 13_0
+              Use `unittest.main()` instead with automatic test discovery.
```
r-owen: I'm not entirely sure what "with automatic test discovery" means. At the very least consider a comma after "instead", but it may not help much.

python/lsst/utils/tests.py
```diff
@@ -197,7 +238,7 @@ class ExecutablesTestCase(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         """Abort testing if automated test creation was enabled and
-        yet not tests were found."""
+        yet no tests were found."""
```
r-owen: Good fix. Consider deleting "yet" as well

python/lsst/utils/tests.py
```diff
@@ -217,17 +258,20 @@ def assertExecutable(self, executable, root_dir=None, args=None, msg=None):
...
-                absolute path.
-
-        root_dir : `str`
-            Directory containing exe. Ignored if None.
-
-        args : list or tuple
+            Path to an executable. ``root_dir`` is not used if this is an
+            absolute path.
+        root_dir : `str`, optional
+            Directory containing exe. Ignored if `None`.
```
r-owen: "exe" -> "executable"; I realize this is pre-existing, but I find it very confusing, since file names for our executables do not end with `.exe`

python/lsst/utils/tests.py
```diff
@@ -262,16 +306,12 @@ def _build_test_method(cls, executable, root_dir):
         The method is built for the supplied excutable located
         in the supplied root directory.
```
r-owen: I realize this is pre-existing, but... do you have any idea what means? I don't.

tests/test_executables.py
```diff
@@ -1,7 +1,35 @@
...
+
+    def testSimpleExe(self):
+        """Test explicit shell script"""
+        self.assertExecutable("testexe.sh",
+                              root_dir=os.path.dirname(__file__),
+                              args=["-h"],
+                              msg="testexe.sh failed")
+
+        # Try a non-existent test
+        with self.assertRaises(Exception):
```
r-owen: Can you safely be more specific than `Exception`? Perhaps `AssertionError`? Similarly in several places below

tests/test_get_caller_name.py
```diff
@@ -31,7 +31,8 @@ class GetCallerNameTestCase(unittest.TestCase):
     """Test get_caller_name
 
     Warning: due to the different ways this can be run
-    (e.g. directly or py.test, the module name can be one of two different things)
+    (e.g. directly or py.test, the module name can be one of two different
+    things)
```
r-owen: Pre-existing, but please move the `)` to just after `py.test`

python/lsst/utils/tests.py
timj: Reworded.

python/lsst/utils/tests.py
timj: clarification: file descriptor leak tester

python/lsst/utils/tests.py
timj: For each executable I create a new method that runs that executable. I attach that method to the test class so that pytest will find it and run the test method later.

PR #55, opened 2018-07-17
DM-14828: Update pybind11 wrappers for 2.2

r-owen:


tests/cache.cc
```diff
@@ -1,8 +1,5 @@
 #include "lsst/utils/python/Cache.h"
 
-PYBIND11_PLUGIN(_cache) {
-    py::module mod("_cache");
+PYBIND11_MODULE(_cache, mod) {
     lsst::utils::Cache<int, std::string> cache;
-    lsst::utils::python::declareCache<int, std::string>(mod, "NumbersCache");
-    return mod.ptr();
-}
+    lsst::utils::python::declareCache<int, std::string>(mod, "NumbersCache");}
```
natelust: Normally the } is on its own line, why not do that here?

r-owen: It was an unwanted side effect of my automatic replacement. I fixed it.

PR #58, opened 2018-08-30
DM-15599: Stop using boost persistence in afw

r-owen:
I am not sure why this change was required.


python/lsst/utils/wrappers.py
```diff
@@ -377,6 +377,8 @@ def register(self, key, subclass):
                     isStatic = isinstance(obj, staticmethod)
                     if isBuiltin or isStatic:
                         if hasattr(self, name):
+                            if name != "__init__subclass__":
+                                continue
```
TallJimbo: I'd like to avoid this special-casing if we can.  Can you say anything more about the conditions under which it was needed?

r-owen: I moved it to the `if name in "__new__":` test, and that does appear to be a new method we can expect on objects. Please have another look.

PR #59, opened 2018-09-20
DM-15809: Replace boost::regex with std::regex

timj:


src/RaDecStr.cc
```diff
@@ -184,8 +184,8 @@ double sexagesimalStrToDecimal(
 
     //Search for leading - sign.
     std::string pmStr = "^-";
-    static const boost::regex pm(pmStr);
-    if(boost::regex_search(inStr, pm))
+    static const std::regex pm(pmStr);
```
ktlim: This seems a bit excessive for checking the first character.

src/RaDecStr.cc
```diff
@@ -164,11 +164,11 @@ double sexagesimalStrToDecimal(
...
-    boost::cmatch what;
+    const std::regex re(regexStr);
+    std::cmatch what;
     //This throws an exception of failure. I could catch it,
     //but I'd only throw it again
-    if(! boost::regex_search(inStr.c_str(), what, re)) {
+    if(! std::regex_search(inStr.c_str(), what, re)) {
         std::string msg= boost::str(boost::format(
             "Failed to parse %s as a right ascension or declination with regex %s")
             % inStr % regexStr);
```
ktlim: Below, I believe `std::cmatch` supports `what[1].str()` to simplify the conversions.

timj:
#62 is going to supersede this PR to a certain extent. I'll revisit it once the radec stuff has been removed.

PR #60, opened 2018-09-21
DM-10384: Add tooling for single-module-per-package pybind11

TallJimbo:


include/lsst/utils/python.h
```diff
@@ -155,6 +160,296 @@ inline std::pair<std::size_t, std::size_t> cppIndex(std::ptrdiff_t size_i, std::
...
+     * `finish()` should be called exactly once, at or near the end of a
+     * `PYBIND11_MODULE` block.
+     */
+    void finish() {
+        for (auto dep = _dependencies.begin(); dep != _dependencies.end(); dep = _dependencies.erase(dep)) {
+            pybind11::module::import(dep->c_str());
+        }
+        for (auto def = _definitions.begin(); def != _definitions.end(); def = _definitions.erase(def)) {
+            (def->second)(def->first);  // WrapperCallback(module)
+        }
```
parejkoj: Do we need to prevent, and is there any way to check, whether `finish` was called multiple times?

include/lsst/utils/python.h
```diff
@@ -155,6 +160,296 @@ inline std::pair<std::size_t, std::size_t> cppIndex(std::ptrdiff_t size_i, std::
...
+    }
+
+    /**
+     * The module object passed to the `PYBIND11_MODULE` block that contains
+     * this WrapperCollection.
+     */
+    pybind11::module module;
```
parejkoj: I'm wondering why this one is public, but the rest not?

tests/example.cc
```diff
@@ -81,48 +81,60 @@ bool acceptNumberConstRef(std::numeric_limits<T> *) { return false; }
...
-    mod.def("accept_cref_int8", [](signed char const &val){return acceptNumber(val);});
-    mod.def("accept_cref_int16", [](short const &val){return acceptNumber(val);});
-    mod.def("accept_cref_int32", [](int const &val){return acceptNumber(val);});
-    mod.def("accept_cref_int64", [](long long const &val){return acceptNumber(val);});
-
-    mod.def("getName", (std::string (*)(int)) getName);
-    mod.def("getName", (std::string (*)(double)) getName);
-}
\ No newline at end of file
+    lsst::utils::python::WrapperCollection wrappers(mod, "lsst.geom");
```
parejkoj: Why is `lsst.geom` included in this example? It doesn't seem to be needed.

python/lsst/utils/wrappers.py
```diff
@@ -79,6 +79,12 @@ class Foo:
             def run(self):
                 return None
 
+    .. warning::
+
+        Python's built-in `super` function does not behave properly in a
+        `continueClass` block.  Base class methods must be invoked directly
```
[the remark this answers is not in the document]
TallJimbo: Updated to clarify: `super` "does not behave properly when used in a class decorated with `continueClass`.

include/lsst/utils/python.h
```diff
@@ -155,6 +160,296 @@ inline std::pair<std::size_t, std::size_t> cppIndex(std::ptrdiff_t size_i, std::
...
+ * their associated methods to be declared at a single point, but the method
+ * wrappers wrapped in a lambda to defer their execution.  A single
+ * WrapperCollection instance is typically constructed at the beginning of
+ * a PYBIND11_MODULE block, then passed by reference to wrapper-declaring
+ * functions defined in other source files.  As type and method wrappers are
+ * added to the WrapperCollection by those functions, the types are registered
+ * immediately, and the method-wrapping lambdas are collected.  After all
+ * wrapper-declaring functions have been called, finish() is called at the
+ * end of the PYBIND11_MODULE block to execute the collecting method-wrapping
+ * lambdas.
```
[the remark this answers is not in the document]
TallJimbo: Thanks!  The devguide covers essentially the same stuff, but is intended more as a step-by-step tutorial for beginners (though it also goes into more detail in places), while this is intended more as a quick reference for people who maybe did this once before but forgot the details.

include/lsst/utils/python.h
```diff
@@ -155,6 +160,296 @@ inline std::pair<std::size_t, std::size_t> cppIndex(std::ptrdiff_t size_i, std::
...
+ *             mod.def("freeFunction", &freeFunction);
+ *         }
+ *     );
+ * }
+ * @endcode
+ *
+ * Note that we recommend the use of universal lambdas (i.e. `auto &`
+ * parameters) to reduce verbosity.
+ */
+class LSST_PRIVATE WrapperCollection final {
```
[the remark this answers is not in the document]
TallJimbo: I've added a code comment.  It's basically telling the linker that it can strip out this symbol from the binary, since nothing outside the module will ever call it.  That actually does nothing for our compiler settings, but it's good practice to reduce binary sizes, link load times, and particularly linker warnings.

include/lsst/utils/python.h
```diff
@@ -155,6 +160,296 @@ inline std::pair<std::size_t, std::size_t> cppIndex(std::ptrdiff_t size_i, std::
...
+     *
+     * @return the `cls` argument for convenience
+     */
+    template <typename PyType, typename ClassWrapperCallback>
+    PyType wrapType(PyType cls, ClassWrapperCallback function, bool setModuleName=true) {
+        if (setModuleName) {
+            cls.attr("__module__") = _package;
+        }
+        wrapFunctions(
+            [cls=cls, function=std::move(function)] (pybind11::module & mod) mutable -> void {
```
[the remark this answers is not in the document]
TallJimbo: It's mutable because it wants to make changes to `cls` rather than just use it as a `const` variable.  I'll add a code comment.

include/lsst/utils/python.h
```diff
@@ -155,6 +160,296 @@ inline std::pair<std::size_t, std::size_t> cppIndex(std::ptrdiff_t size_i, std::
...
+     */
+    void wrapFunctions(WrapperCallback function) {
+        _definitions.emplace_back(std::make_pair(module, function));
+    }
+
+    /**
+     * Add a type (class or enum) wrapper, deferring method and other
+     * attribute definitions until finish() is called.
+     *
+     * @param[in] cls       A `pybind11::class_` or `enum_` instance.
```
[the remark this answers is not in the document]
TallJimbo: I think it's best to document template params if and only if they need to be specified explicitly (as in class templates, or function templates where only the return type is templated).  This particular method should only be called with arguments that let the compiler infer the template parameters, which IMO makes them something of an implementation detail.

include/lsst/utils/python.h
TallJimbo: Because I'm emptying the lists as I process them, a second call to `finish()` should be a no-op.  I do think it'd be slightly cleaner interface to prevent that, but the implementation of that is enough messier that it's probably best overall as it is.

include/lsst/utils/python.h
TallJimbo: It's public to save people writing `wrapSomething(WrapperCollection & wrappers)` functions from having to always pass a separate module argument in case they might need it (note that you need a module for the `py::class_` invocations that happen in the first argument to `wrapType`.  I could have made it private with a getter method, but that feels a bit like just pretending that something is private when it really isn't.

tests/example.cc
TallJimbo: I was just using it as an example namespace to check later.  I'll use a different string that's more obvious about that.

include/lsst/utils/python.h
```diff
@@ -155,6 +160,300 @@ inline std::pair<std::size_t, std::size_t> cppIndex(std::ptrdiff_t size_i, std::
...
+ * // _mypackage.cc
+ *
+ * void wrapClassA(WrapperCollection & wrappers);
+ * void wrapClassB(WrapperCollection & wrappers);
+ *
+ * PYBIND11_MODULE(_mypackage, module) {
+ *     WrapperCollection wrappers(module, "mypackage");
+ *     wrapClassA(wrappers);
+ *     wrapClassB(wrappers);
+ *     wrappers.finish();
```
pschella: Why not have this done by RAII?

pschella: Depends on what developers find more surprising I guess, forgetting to call `finish` and have no wrappers or having it happen upon destruction. Also what happens if `finish` is called twice?

include/lsst/utils/python.h
```diff
@@ -155,6 +160,300 @@ inline std::pair<std::size_t, std::size_t> cppIndex(std::ptrdiff_t size_i, std::
...
+    WrapperCollection & operator=(WrapperCollection const &) = delete;
+    WrapperCollection & operator=(WrapperCollection &&) = delete;
+
+    ~WrapperCollection() noexcept {
+        if (!std::uncaught_exception() && !_definitions.empty()) {
+            PyErr_SetString(PyExc_ImportError,
+                            "WrapperCollection::finish() not called; module definition incomplete.");
+            PyErr_WriteUnraisable(module.ptr());
+        }
+    }
```
pschella: Is this more or less surprising?

PR #61, opened 2018-09-22
DM-15836: Add helper class to make wrapping templated functions easier.

TallJimbo:


TallJimbo:
One possibility that I'm pondering now that I'm using this code in practice down in afw: instead of calling a named `fail` method when none of the given types match, should I call `operator()` with some sentinal type instead?  That would make it possible to use a universal lambda as the callable, though you'd to play `decltype` tricks inside it to extract the `Tag` type from the argument, and I'm not sure you can handle the failure case well in the same body without C++17's `constexpr if`.  In afw right now, I've written a single trampoline class to pass to `TemplateInvoker` that is itself templated on another callable that *can* be a universal lambda; that works because I have to wrap a lot of methods with the same `fail` but different behavior when the match succeeds.

include/lsst/utils/python/TemplateInvoker.h
```diff
@@ -0,0 +1,151 @@
...
+ * );
+ * @endcode
+ *
+ * The `fail()` method should always return `py::object`, even if it throws
+ * (though if it always throws it should probably be tagged with the
+ * `[[noreturn]]` attribute).
+ *
+ * The type returned by the helper callable's `operator()` can be anything
+ * pybind11 knows how to convert to Python.
+ */
```
natelust: Consider changes as spoken in person

PR #63, opened 2018-09-26
DM-15871: add smart doImport function from daf_butler

TallJimbo:


tests/test_doImport.py
```diff
@@ -0,0 +1,67 @@
...
+        self.assertTrue(inspect.isfunction(c))
+
+        c = doImport("lsst.utils")
+        self.assertTrue(inspect.ismodule(c))
+
+        with self.assertRaises(ImportError):
+            doImport("lsst.utils.tests.TestCase.xyprint")
+
+        with self.assertRaises(ImportError):
+            doImport("lsst.daf.butler.nothere")
```
timj: Maybe don't use a butler example here (just in case something weird happens in butler out of your control).

PR #50, opened 2018-03-23
DM-13854: add general-purpose Cache class

PaulPrice:

PR #65, opened 2018-10-29
DM-9938: Make some afw types hashable

kfindeisen:
This PR adds several extensions to the hash support added in #50.

I've added several functions for testing compliance with the `std::hash` spec. I have *not* added a hash uniformity test. I found that GCC's implementation of `std::hash<double>` fails even a very naive bitwise test, and I don't want to require that implementers of classes whose state is a single primitive jump through hoops to meet a higher standard.

### DIFF
--- a/include/lsst/utils/hashCombine.h
+++ b/include/lsst/utils/hashCombine.h
@@ -66,6 +66,38 @@ std::size_t hashCombine(std::size_t seed, const T& value, Rest... rest) noexcept
     return hashCombine(seed, rest...);
 }
 
+/**
+ * Combine hashes in an iterable.
+ *
+ * This is provided as a convenience for those who need to hash a container.
+ *
+ * @tparam InputIterator an iterator to the objects to be hashed. The
+ *                       pointed-to type must have a valid (in particular,
+ *                       non-throwing) specialization of std::hash.
+ *
+ * @param seed An arbitrary starting value.
+ * @param begin, end The range to hash.
+ * @returns A combined hash for all the elements in [begin, end).
+ *
+ * @exceptsafe Shall not throw exceptions.
+ *
+ * To use it:
+ *
+ *     // Arbitrary seed; can change to get different hashes of same argument list
+ *     std::size_t seed = 0;
+ *     result = hashIterable(seed, container.begin(), container.end());
+ */
+// Note: not an overload of hashCombine to avoid ambiguity with hashCombine(size_t, T1, T2)
+// WARNING: should not be inline or constexpr; it can cause instantiation-order problems with std::hash<T>
+template <typename InputIterator>
+std::size_t hashIterable(std::size_t seed, InputIterator begin, InputIterator end) noexcept {
+    std::size_t result = 0;
+    for (; begin != end; ++begin) {
+        result = hashCombine(result, *begin);
+    }
+    return result;
+}
+
 }} // namespace lsst::utils
 
 #endif
--- a/include/lsst/utils/hashCombine.h
+++ b/include/lsst/utils/hashCombine.h
@@ -27,30 +27,44 @@
 namespace lsst {
 namespace utils {
 
-//@{
-/** Combine hashes
+/**
+ * Combine hashes
+ *
+ * A specialization of hashCombine for a trivial argument list.
+ */
+inline std::size_t hashCombine(std::size_t seed) noexcept { return seed; }
+
+/**
+ * Combine hashes
  *
  * This is provided as a convenience for those who need to hash a composite.
  * C++11 includes std::hash, but neglects to include a facility for
  * combining hashes.
  *
+ * @tparam T, Rest the types to hash. All types must have a valid (in
+ *                 particular, non-throwing) specialization of std::hash.
+ *
+ * @param seed An arbitrary starting value.
+ * @param value, rest The objects to hash.
+ * @returns A combined hash for all the arguments after `seed`.
+ *
+ * @exceptsafe Shall not throw exceptions.
+ *
  * To use it:
  *
- *    std::size_t seed = 0;
- *    result = hashCombine(seed, obj1, obj2, obj3);
- *
- * This solution is provided by Matteo Italia
- * https://stackoverflow.com/a/38140932/834250
+ *     // Arbitrary seed; can change to get different hashes of same argument list
+ *     std::size_t seed = 0;
+ *     result = hashCombine(seed, obj1, obj2, obj3);
  */
-inline std::size_t hashCombine(std::size_t seed) { return seed; }
-
+// This implementation is provided by Matteo Italia, https://stackoverflow.com/a/38140932/834250
+// Algorithm described at https://stackoverflow.com/a/27952689
+// WARNING: should not be inline or constexpr; it can cause instantiation-order problems with std::hash<T>
 template <typename T, typename... Rest>
-std::size_t hashCombine(std::size_t seed, const T& value, Rest... rest) {
+std::size_t hashCombine(std::size_t seed, const T& value, Rest... rest) noexcept {
     std::hash<T> hasher;
     seed ^= hasher(value) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
     return hashCombine(seed, rest...);
 }
-//@}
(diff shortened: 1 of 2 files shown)


include/lsst/utils/hashCombine.h
```diff
@@ -27,30 +27,72 @@
...
- *    std::size_t seed = 0;
- *    result = hashCombine(seed, obj1, obj2, obj3);
+ * @exceptsafe Shall not throw exceptions.
  *
- * This solution is provided by Matteo Italia
- * https://stackoverflow.com/a/38140932/834250
+ * To use it:
+ *
+ *     std::size_t seed = 0;
+ *     result = hashCombine(seed, obj1, obj2, obj3);
```
TallJimbo: If `seed` shouldn't always be zero (and if it were, I assume that'd be done internally)`, it'd be good to provide some guidance as to what it should be in typical implementations of `std::hash` that use it.

include/lsst/utils/hashCombine.h
```diff
@@ -27,30 +27,72 @@
...
  * This is provided as a convenience for those who need to hash a composite.
  * C++11 includes std::hash, but neglects to include a facility for
  * combining hashes.
  *
- * To use it:
+ * @tparam T, Rest the types to hash. All types must have a valid (in
+ *                 particular, non-throwing) specialization of std::hash.
+ *
+ * @param seed An arbitrary starting value.
+ * @param value, rest The objects to hash.
```
TallJimbo: I'm open to persuasion on this, but I'm not convinced it's valuable to document template parameters (via `tparam`) separately from regular arguments (via `param`) when the template parameters are expected to be completely inferred rather than explicitly provided.  In those conditions, I feel like the actual template parameters are something of an implementation detail (albeit an unfortunately visible one, given the constraints of the language), and the documentation works better when the regular `param` docs just include information about permitted types.

include/lsst/utils/tests.h
```diff
@@ -0,0 +1,110 @@
...
+
+#include <ostream>
+#include <type_traits>
+
+namespace lsst {
+namespace utils {
+
+namespace {
+// Variable template reporting whether a type can be printed using <<
+template <typename, typename = void>
```
TallJimbo: Took me a moment to figure out why you needed two template parameters here...might be worth a comment.

include/lsst/utils/tests.h
```diff
@@ -0,0 +1,110 @@
...
+constexpr bool HAS_STREAM_OUTPUT = false;
+template <typename T>
+constexpr bool HAS_STREAM_OUTPUT<
+        T, std::enable_if_t<true, decltype((void)(std::declval<std::ostream&>() << std::declval<T&>()),
+                                           void())>> = true;
```
TallJimbo: Could you walk me through the `decltype` expression here?  I see that you're trying to invoke the `ostream` output operator, which will return `std::ostream &` if it succeeds (and SFINAE out of this specialization if it fails), and you then cast that to `void`, I'm guessing to avoid warnings about unused values while ensuring the compiler doesn't optimize it away?  I wouldn't have thought that would be a problem inside `decltype`, so I suspect I'm confused.  And I was totally lost about why you're invoking the comma *operator* with `void()` as the RHS argument - that would make the `decltype` expression always return `void`, but I don't see why we actually care about what it returns.

include/lsst/utils/tests.h
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Compile-time test of whether a specialization of std::hash conforms to the
+ * general spec.
+ *
+ * The function itself is a no-op.
+ *
+ * @tparam T The properties of `std::hash<T>` will be tested.
+ */
+template <typename T>
+inline void assertValidHash() {
```
TallJimbo: Would it be useful to make this `constexpr`?  Certainly seems like it could be, and perhaps ought to be.

include/lsst/utils/hashCombine.h
kfindeisen: Unfortunately, there isn't much guidance available. This example is from the older documentation (I just moved it to the style-guide-recommended order), where it appears to have been copied verbatim from one of the Stack Overflow discussions. I suppose I could point out that changing the seed lets you make sure two hash functions of similar types are different, or something like that...

Incidentally, I'm not sure 0 is a good choice for a seed because it can lead to degeneracies with some combiner algorithms. This one is safe, but separation of interface and implementation and all that...

include/lsst/utils/hashCombine.h
kfindeisen: I don't see how template parameters are an implementation detail, given that templates are frequently used to represent flexible types rather than arbitrary ones. For example, often two parameters are required to be of the same type, or the output type and/or validity of future calls depends on the parameter. The programmer needs to be aware of such constraints.

Even in cases like this one where the parameters are *almost* unconstrained, it seems useful to separate type from value information. While you certainly could merge all the information from a type parameter description into the corresponding parameter documentation, that often leads to a single very long and hard to read paragraph.

As for this being a C++ quirk, Java, which has a completely different implementation of generic programming, [also documents type parameters](https://docs.oracle.com/javase/10/docs/api/java/util/Map.html).

include/lsst/utils/tests.h
kfindeisen: Unfortunately, this bit of code has been cargo-culted (believe me, I tried to find *anything* that was more readable). I think I combined the second example from the [`void_t` docs](https://en.cppreference.com/w/cpp/types/void_t) with a more general metaprogramming example. I'm guessing the evaluated type needs to be `void` to ensure the compiler doesn't try to use it in the template proper, but otherwise I understand about as much as you do.

include/lsst/utils/hashCombine.h
TallJimbo: I'm by no means against documenting template parameters separately in general, and not very opposed to it at all in this case; I just think it's worth thinking about, especially in the case of variadics and overloads.

I've also been thinking about whether the case where template parameters must be explicitly provided (e.g. the first argument to `std::make_shared`) is sufficiently different from the case where they're inferred (all other `make_shared` arguments) that we should reserving `@tparam` to draw attention to the former cases - but that's a general question about Doxygen usage, not a comment on this ticket, and I suppose it's something Doxygen would probably warn about anyway.

Anyhow, as an example of what I mean by "implementation detail", one could imagine implementing this API as:
```
template <typename ...Rest>
void hashCombine(int seed, Rest...rest);
```
...if, for some reason, we were forwarding to some other implementation that separated the first template argument from the rest of the parameter pack.
From the caller's perspective, that wouldn't shouldn't change the API, unless they were depending on the detailed signature in a way that I think we'd both agree is inadvisable.

Anyhow, I don't think it's a problem to document the function as it's written, especially because the signature you did use here does reflect what the function requires (and my counterfactual one does not!) and there are no non-template overloads that we'd like to maintain consistency with.

include/lsst/utils/hashCombine.h
```diff
@@ -27,30 +27,72 @@
...
+ *     std::size_t seed = 0;
+ *     result = hashCombine(seed, obj1, obj2, obj3);
  */
-inline std::size_t hashCombine(std::size_t seed) { return seed; }
-
+// This implementation is provided by Matteo Italia, https://stackoverflow.com/a/38140932/834250
+// Algorithm described at https://stackoverflow.com/a/27952689
 template <typename T, typename... Rest>
-std::size_t hashCombine(std::size_t seed, const T& value, Rest... rest) {
+std::size_t hashCombine(std::size_t seed, const T& value, Rest... rest) noexcept {
```
TallJimbo: Didn't spot this at first, but should this be `Rest const &... rest` (or `Rest &&... rest` with `std::forward`)?

kfindeisen: I don't know. I don't see why the former would hurt, given that the `const T&` and the recursive nature of the function impose that constraint on all arguments anyway (universal references, on the other hand, have too many gotchas for my liking). However, none of the [cppreference examples](https://en.cppreference.com/w/cpp/language/parameter_pack) use anything but a straight parameter name, so I think a case can be made that this is at least more idiomatic.

PR #67, opened 2019-02-04
DM-17534: Create generic iterable flattener

kfindeisen:
This PR adds a general purpose flattening function for nested iterables, including support for multiple layers of nesting and (the original requirement) mixtures of scalar and iterable elements.

### DIFF
--- a/python/lsst/utils/iterable.py
+++ b/python/lsst/utils/iterable.py
@@ -1,0 +1,70 @@
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+__all__ = ["flatten"]
+
+from collections.abc import Mapping
+
+
+def flatten(nested):
+    """Flatten an iterable of possibly nested iterables.
+
+    Parameters
+    ----------
+    nested : iterable
+        An iterable that may contain a mix of scalars or other iterables.
+
+    Returns
+    -------
+    flat : sequence
+        A sequence where each iterable element of `nested` has been replaced
+        with its elements, in order, and so on recursively. The result
+        preserves the iteration order of the inputs.
+
+    Raises
+    ------
+    ValueError
+        Raised if `nested` is or contains a mapping type (such as `dict`).
+        While they are iterable, mapping types cannot be converted into
+        sequences or subsequences in any reasonable way.
+
+    Examples
+    --------
+    >>> x = [42, [4, 3, 5]]
+    >>> flatten(x)
+    [42, 4, 3, 5]
+    """
+    if isinstance(nested, Mapping):
+        raise ValueError("Can't flatten mapping: %s" % nested)
+
+    flat = []
+    for x in nested:
+        if isinstance(x, (str, bytes)):
+            flat.append(x)
+        else:
+            # Best way to test for an iterable
+            # see https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable
+            try:
+                iter(x)
+                flat.extend(flatten(x))
+            except TypeError:
+                flat.append(x)
+    return flat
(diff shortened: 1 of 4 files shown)


python/lsst/utils/iterable.py
```diff
@@ -0,0 +1,53 @@
...
+    -------
+    flat : sequence
+        A sequence where each iterable element of `nested` has been replaced
+        with its elements, in order, and so on recursively. The result
+        preserves the iteration order of the inputs.
+
+    Examples
+    --------
+    >>> x = [42, [4, 3, 5]]
+    >>> _flatten(x)
```
ktlim: No underscore here, right?

python/lsst/utils/iterable.py
```diff
@@ -0,0 +1,53 @@
...
+    [42, 4, 3, 5]
+    """
+    flat = []
+    for x in nested:
+        try:
+            iter(x)
```
ktlim: This guarantees an exception per element in `flat`.  Is this sufficiently efficient for a low-level routine in `utils`?

python/lsst/utils/iterable.py
```diff
@@ -0,0 +1,53 @@
...
+    ----------
+    nested : iterable
+        An iterable that may contain a mix of scalars or other iterables.
+
+    Returns
+    -------
+    flat : sequence
+        A sequence where each iterable element of `nested` has been replaced
+        with its elements, in order, and so on recursively. The result
+        preserves the iteration order of the inputs.
```
ktlim: Mentioning what happens to unordered sets might be good.

python/lsst/utils/iterable.py
```diff
@@ -0,0 +1,53 @@
...
+    --------
+    >>> x = [42, [4, 3, 5]]
+    >>> _flatten(x)
+    [42, 4, 3, 5]
+    """
+    flat = []
+    for x in nested:
+        try:
+            iter(x)
+            flat.extend(flatten(x))
```
ktlim: If you pass `flatten()` a string, which is an iterable, its components are also strings, which are still iterable, even when a single character.  This recursion then never terminates.
This also happens if you pass `flatten()` a dict (also an iterable) with string keys.

python/lsst/utils/iterable.py
parejkoj: "The result preserves the iteration order of the inputs"

That makes it pretty clear to me: you get whatever the `set` gives you when you iterate on it.

python/lsst/utils/iterable.py
parejkoj: KT is correct here (just tested it): I think you may have to special-case `str` somehow. Possibly a `if isinstance(x, (str, bytes)): flat.append(x)` block inside?

Stackoverflow for reference: https://stackoverflow.com/questions/2158395/flatten-an-irregular-list-of-lists

TallJimbo:
FYI, some combination of @timj and @pschella (forget who) wrote something like this for daf_butler, that could be pulled into somewhere more general and/or improved.

https://github.com/lsst/daf_butler/blob/master/python/lsst/daf/butler/core/utils.py#L32

timj:
See also https://github.com/lsst/daf_persistence/blob/master/python/lsst/daf/persistence/utils.py which also has an equivalent function.

tests/test_iterable.py
```diff
@@ -0,0 +1,59 @@
...
+                subsequenceFound = True
+                break
+        self.assertTrue(subsequenceFound)
+
+    def testPartialOrderingOuter(self):
+        initial = [0, {1, 2}, 3]
+        flat = flatten(initial)
+        self.assertEqual(set(flat), set(range(4)))
+        self.assertEqual(flat[0], 0)
+        self.assertEqual(flat[3], 3)
```
parejkoj: Ugh, non-order preserving-ness does make this hard. If we were fully on py3.7 it would be less of an issue.

parejkoj:
@ktlim : I requested that this be moved to `utils`. Where else would you suggest we put it? It's the sort of thing that I wish the python devs had put in `itertools` or something in the first place.

timj:
If it's used in the original location, daf_butler and daf_persistence (at least) then it probably can justify being generic. On the other hand I am desperately trying to limit daf_butler's dependencies and actively preventing lsst code dependencies wherever possible.

python/lsst/utils/iterable.py
kfindeisen: Hmm... this implies that `doctest` doesn't run in `utils`.

python/lsst/utils/iterable.py
kfindeisen: According to the web, this is the *only* sure way to test if a variable is iterable. (Presence of an `__iter__` method and inheritance from `collections.abc.Iterable` are each sufficient but not necessary conditions).

kfindeisen:
I'm not sure the existing code noted above by @TallJimbo and @timj is relevant: both examples are for converting scalars to iterables, which is almost the opposite of what `flatten` does. I take it `flatten` isn't worth putting in `utils` if it can't replace the `daf_butler` and `daf_persistence` cases?

kfindeisen:
Closing this PR until we find a second client for `flatten`, at which point sharing the code will make much more sense.

PR #76, opened 2019-11-12
DM-22199: Add decorators for iterating over tests

PaulPrice:


python/lsst/utils/tests.py
```diff
@@ -707,3 +708,111 @@ def assertFloatsEqual(testCase, lhs, rhs, **kwargs):
...
+        If the ``settings`` are not of the same length.
+
+    Yields
+    ------
+    parameters : `dict` (`str`: anything)
+        Set of parameters.
+    """
+    num = len(next(iter(settings.values())))  # Number of settings
+    for name, values in settings.items():
+        if isinstance(values, str):
```
ktlim: I think you have to do this as a separate loop, first.  If the first item in `settings` is a string, `num` will get set to the length of that string instead of 1 and will then trigger the assertion.

PaulPrice: Well spotted! I've pushed a fix.

PR #77, opened 2019-11-13
DM-22207:Update doc string to specify this works on classes as well as functio…

leannep:


python/lsst/utils/deprecated.py
```diff
@@ -24,7 +24,7 @@
 
 
 def deprecate_pybind11(func, reason, category=FutureWarning):
-    """Deprecate a pybind11-wrapped C++ interface function or method.
+    """Deprecate a pybind11-wrapped C++ interface function, method or class.
```
jonathansick: Adding a serial comma is great here.

PR #91, opened 2020-11-19
DM-27638: Try to ensure that previous module not found error is included

timj:
Since doImport treats ModuleNotFoundError as a special error
indicating that it should probably retry with a different
subset of the supplied string, this means that failures
to import dependencies trigger a confusing error report.
It's hard to disentangle real errors from dependencies
versus a request for a class within a module, so as a short
term fix include the previous import error in the final
error. Hopefully this will mitigate the problem.

Also add some explicit tests for this scenario.


MichelleGower (approved):
Code changes looked good.  Tested it and got the more helpful error message.

PR #94, opened 2021-05-04
DM-30023: replace uncaught_exception by uncaught_exceptions

mwittgen:
From C++17 on bool uncaught_exception is deprecated in favor of int uncaught_exceptions


erykoff (approved):
Looks good to me! (Checking https://en.cppreference.com/w/cpp/error/uncaught_exception for reference).

PR #96, opened 2021-07-20
DM-31141: Add test decorators for cartesian product

PaulPrice:


python/lsst/utils/tests.py
```diff
@@ -842,6 +843,112 @@ def wrapper(self, *args, **kwargs):
     return decorator
 
 
+def cartesianProduct(settings):
```
timj, line 846: I assume this is a private function that should start with an underscore?

python/lsst/utils/tests.py
```diff
@@ -842,6 +843,112 @@ def wrapper(self, *args, **kwargs):
...
+
+        cartesianProduct({"foo": [1, 2], "bar": ["black", "white"]})
+
+    returns:
+
+        {"foo": [1, 1, 2, 2], "bar": ["black", "white", "black", "white"]}
+    """
+    product = {kk: [] for kk in settings}
+    for values in itertools.product(*settings.values()):
+        for kk, vv in zip(settings.keys(), values):
```
timj, line 871: Is the `keys()` necessary there?

tests/test_decorators.py
```diff
@@ -53,13 +55,44 @@ def teardown_method(self, method):
...
+        self.assertEqual(self.__class__.__name__, f"DecoratorProductTestCase_{self.word}_{self.number}")
+
+    @lsst.utils.tests.methodParametersProduct(
+        xx=[1, 2, 3],
+        yy=[9, 8],
+    )
+    def testMethodDecorator(self, xx, yy):
+        self.combinations.add((xx, yy))
+
+    def teardown_method(self, method):
```
timj, line 77: is this pytest-specific?

timj (approved):
Looks okay.

python/lsst/utils/tests.py
PaulPrice: I wouldn't have called it private: it doesn't have any hidden implementation details. It's just a helper function. It doesn't get exported in the `__all__` only because I didn't think it would be useful to anyone, but does that make it "private"?

timj: Okay. I tend to put an `_` in front of functions as an additional "don't rely on this existing in the future" hint since people can pull functions out of packages even if they aren't in `__all__`. Up to you.

tests/test_decorators.py
PaulPrice: Ah, it seems it is -- it doesn't fire when run with `python tests/test_decorators.py`. I copied the earlier implementation, which also uses it.
Do I need to find an alternative implementation?

timj: It's somewhat ironic that the test of this new decorator that implements functionality already present in pytest, requires pytest to be tested.  The test passes when run as `python tests/...` or it fails?

PaulPrice: It passes. It just doesn't hit this additional check, and so isn't exercising the full test.
I've got an alternative implementation in mind. I'll give it a go.

PaulPrice: OK, lemme know what you think of this.

timj: Using a test flag and `tearDown` is fine with me.

python/lsst/utils/tests.py
timj: To clarify my comment, I don't think `.keys()` is needed because `zip` will automatically get the keys from the dict.

```
>>> a = {1:2, 3:4}
>>> b = {"a":"b", "c":"d"}
>>> print(list(zip(a, b)))
[(1, 'a'), (3, 'c')]
```

PaulPrice: Absolutely. I updated the code already.

timj: 👍 I realized my comment was ambiguous such that the thumbs up from you was ambiguous to me 😄

PaulPrice:
OK, I think everyone is happy. Merge is imminent.

PR #99, opened 2021-09-23
DM-31046: Add mime.cache to file leaks ignore list

parejkoj:
I'm adding code to analysis_ap that uses PIL to open PNG images, which can leave the `mime.cache` open. We don't have control over that, so just ignore file leak failures due to this.


kfindeisen:
The change itself looks fine; since I'm not familiar with PIL, I can't comment on whether there's a way to avoid using the cache or ensure it's promptly closed.

PR #100, opened 2021-09-28
DM-31722: Migrate utility code from other packages

timj:
Code is coming from pipe_base and daf_butler.

## Checklist

- [ ] ran Jenkins


python/lsst/utils/timer.py
```diff
@@ -0,0 +1,311 @@
...
+                logger = obj.log
+            except AttributeError:
+                pass
+    strList = []
+    for name, value in pairs:
+        if metadata is not None:
+            _add_to_metadata(metadata, name, value)
+        strList.append(f"{name}={value}")
+    if logger is not None:
+        logging.getLogger("timer." + logger.name).log(logLevel, "; ".join(strList))
```
timj, line 118: Should we be trying to adjust stacklevel for this `logger.log` call so that the log message appears to be from the caller? The issue then becomes whether it should walk the stack to find the first frame that is not in this file or if logPairs should take a stacklevel so that logInfo will know that it wants its parent to be used.

python/lsst/utils/timer.py
```diff
@@ -0,0 +1,311 @@
...
+            msg = ""
+
+        if not success:
+            # Something went wrong so change the log level to indicate
+            # this.
+            level = logging.ERROR
+
+        # Specify stacklevel to ensure the message is reported from the
+        # caller (1 is this file, 2 is contextlib, 3 is user)
+        log.log(level, msg + "%sTook %.4f seconds", *args,
```
timj, line 310: Do we want this to call `logPairs` or `logInfo`? If we do, then we'd need to change things so that the duration information can be reported through `logInfo`. It seems preferable to report the duration and not just leave it for the person analyzing the metadata file to work it out.

python/lsst/utils/logging.py
```diff
@@ -0,0 +1,261 @@
...
+
+# Verbose logging is midway between INFO and DEBUG.
+VERBOSE = (logging.INFO + logging.DEBUG) // 2
+logging.addLevelName(VERBOSE, "VERBOSE")
+
+
+class _F:
+    """
+    Format, supporting str.format() syntax
+    see using-custom-message-objects in logging-cookbook.html
```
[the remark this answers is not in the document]
timj: This is a private class but I'll put a full URL in there.

python/lsst/utils/timer.py
timj: Yes, if you look at the LogRecord it reports the file and line that issued the warning and you can add a `stacklevel=N` to tell it to report it from a caller instead. That's how `time_this` is able to say how long something took as if it was directly being called from the code section being timed.

python/lsst/utils/iteration.py
```diff
@@ -0,0 +1,115 @@
...
+        The contents of a chunk of the iterator as a `tuple`. A tuple is
+        preferred over an iterator since it is more convenient to tell it is
+        empty and the caller knows it can be sized and indexed.
+    """
+    it = iter(data)
+    while (chunk := tuple(itertools.islice(it, chunk_size))):
+        yield chunk
+
+
+def iterable(a: Any) -> Iterable[Any]:
```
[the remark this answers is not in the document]
timj: Renamed to `ensure_iterable`.

PR #103, opened 2021-10-12
DM-32210: Use inspect to determine when we leave the utils module

timj:
Previously using the filename the code could be confused if
there was another lsst/utils directory in the path. Now
using inspect.getmodule the code can no for sure when it
leaves the lsst.utils module.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/utils/timer.py
+++ b/python/lsst/utils/timer.py
@@ -22,7 +22,7 @@ import logging
 import resource
 import time
 import datetime
-import traceback
+import inspect
 from contextlib import contextmanager
 
 from typing import (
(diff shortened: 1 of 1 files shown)


python/lsst/utils/timer.py
```diff
@@ -94,9 +94,12 @@ def _find_outside_stacklevel() -> int:
...
     stacklevel = 1  # the default for `Logger.log`
-    stack = traceback.extract_stack()
-    for i, s in enumerate(reversed(stack)):
-        if "lsst/utils" not in s.filename:
+    stack = inspect.stack()
+    for i, s in enumerate(stack):
+        module = inspect.getmodule(s.frame)
+        if module is None:
+            continue
+        if not module.__name__.startswith("lsst.utils"):
```
natelust, line 102: With inspect and frames you need to be very careful about creating local references to frames, as you can create reference cycles, so the fewer local variables you need the better, and cleaning up variables that contain frames before exiting a function is a good idea. This may be overly defensive for how often this function gets used, but it is good practice, especially if someone uses it as a code reference later on.
```suggestion
    for i, s in enumerate(inspect.stack()):
        module = inspect.getmodule(s.frame)
        if module is None:
            continue
        if not module.__name__.statswith("lsst.utils"):
            stacklevel = i
            break
    del s
    return stacklevel
```

timj: This seems highly unpythonic but I'll make the change.

PR #104, opened 2021-10-28
Add cmake support

fritzm:
Adds ability to build package with CMake in addition to eupspkg (allows to be included as submodule in other CMake projects, e.g. Qserv.)

## Checklist

- [x] ran Jenkins

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,4 @@
+cmake_minimum_required(VERSION 3.1)
+project(utils)
+
+add_subdirectory(python/lsst/utils)
--- a/doc/changes/DM-23308.misc.rst
+++ b/doc/changes/DM-23308.misc.rst
@@ -1,0 +1,2 @@
+Add support for building via CMake in addition to eupspkg (allows inclusion as submodule into
+other CMake projects, e.g. Qserv)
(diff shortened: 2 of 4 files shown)


fritzm:
Hmmm...  I tried installing `util` with pip invoked from cmake install.  I ended up with `deprecrated`, `numpy`, `psutil`, and `wrapt`, none of which Qserv either needs or wants in its run container.  Also, the rendered `version.py` was just `0.1.0`, which seems much less useful than the one derived by the cmake on this PR.  That hardly seems an improvement...

If it really seems desirable to avoid adding the cmake files on this PR to utils, perhaps I should just "cherry-pick" the on one def I need to support `log` (`@continueClass`) out of the submodule with some custom shenanigans in Qserv's cmake files?

fritzm:
Oh, looks like I can `--no-deps`, and that's not too bad.  `version.py` is still weak, but looks like that will change soon anyway.  I will run with that and cancel this PR.  Thank you all for the attention here!

timj:
Yes, those are the actual dependencies of the utils package and it's just luck that you are somehow able to import `continueClass` without triggering some of the other imports (I think mostly because `lsst.utils.tests` is not imported by default).  I do indeed have to fix the version problem before I put it on PyPI and I have created a ticket for that since it's also relevant to your cmake version problem.

PR #105, opened 2021-11-19
DM-32658: Only do getmodule if we know the log message will be issued

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/timer.py
+++ b/python/lsst/utils/timer.py
@@ -306,7 +306,7 @@ def timeMethod(_func: Optional[Any] = None, *, metadata: Optional[MutableMapping
 
     def decorator_timer(func: Callable) -> Callable:
         @functools.wraps(func)
-        def wrapper(self: Any, *args: Any, **keyArgs: Any) -> Any:
+        def timeMethod_wrapper(self: Any, *args: Any, **keyArgs: Any) -> Any:
             # Adjust the stacklevel to account for the wrappers.
             # stacklevel 1 would make the log message come from this function
             # but we want it to come from the file that defined the method
@@ -320,7 +320,7 @@ def timeMethod(_func: Optional[Any] = None, *, metadata: Optional[MutableMapping
                 logInfo(obj=self, prefix=func.__name__ + "End", metadata=metadata, logger=logger,
                         logLevel=logLevel, stacklevel=stacklevel)
             return res
-        return wrapper
+        return timeMethod_wrapper
 
     if _func is None:
         return decorator_timer
--- a/python/lsst/utils/timer.py
+++ b/python/lsst/utils/timer.py
@@ -114,7 +114,7 @@ def _find_outside_stacklevel() -> int:
 
 def logPairs(obj: Any, pairs: Collection[Tuple[str, Any]], logLevel: int = logging.DEBUG,
              metadata: Optional[MutableMapping] = None,
-             logger: Optional[logging.Logger] = None) -> None:
+             logger: Optional[logging.Logger] = None, stacklevel: Optional[int] = None) -> None:
     """Log ``(name, value)`` pairs to ``obj.metadata`` and ``obj.log``
 
     Parameters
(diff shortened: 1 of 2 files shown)


timj:
Actually, @ktlim would like me to try the decorator approach to remove the getmodule overhead from runtime.

python/lsst/utils/timer.py
```diff
@@ -156,9 +156,12 @@ def logPairs(obj: Any, pairs: Collection[Tuple[str, Any]], logLevel: int = loggi
...
         # Want the file associated with this log message to be that
-        # of the caller.
-        stacklevel = _find_outside_stacklevel()
-        logging.getLogger("timer." + logger.name).log(logLevel, "; ".join(strList), stacklevel=stacklevel)
+        # of the caller. This is expensive so only do it if we know the
+        # message will be issued.
+        timer_logger = logging.getLogger("timer." + logger.name)
+        if timer_logger.isEnabledFor(logLevel):
+            stacklevel = _find_outside_stacklevel()
+            timer_logger.log(logLevel, "; ".join(strList), stacklevel=stacklevel)
```
erykoff, line 164: Is it possible to do this once and cache it?  Because this is the wrapped function, right?

timj:
I've updated the PR to pass the stacklevel from timeMethod and through logInfo to logPairs. This should remove the need for getmodule completely.

python/lsst/utils/timer.py
timj: cache what? This is the code that issues the log message. You mean cache `timer_logger` (or calculate it in timeMethod itself?). I think in general `getLogger` is not thought of as being a high cost call. The initial logger usually comes from `Task.log` but can come from a parameter to `timeMethod`. Whilst there is a `logger.getChild`  method there isn't an API for prefixing a logger.

erykoff: What I mean is that even in the logging case we still have the fairly expensive `_find_outside_stacklevel()` call.  I'm wondering if this needs to be computed for every log call (can it change?)

timj: Or you mean the stacklevel?

erykoff: Yeah, whatever is returned by `_find_outside_stacklevel()` (though maybe this is much less common with your new changes?)

python/lsst/utils/timer.py
```diff
@@ -281,16 +306,18 @@ def run(self, ...): # or any other instance method you want to time
 
     def decorator_timer(func: Callable) -> Callable:
         @functools.wraps(func)
-        def wrapper(self: Any, *args: Any, **keyArgs: Any) -> Any:
+        def timeMethod_wrapper(self: Any, *args: Any, **keyArgs: Any) -> Any:
+            # Adjust the stacklevel to account for the wrappers.
+            stacklevel = 2
```
TallJimbo, line 311: I have zero confidence that I could guess whether this should be `2` or `1` without testing, so I'm just going to assume you tested it somehow at least once.  We do have unit tests for this, right?

python/lsst/utils/timer.py
TallJimbo: Yeah, I think `_find_outside_stacklevel` is basically never going to be called anymore. I wonder a bit whether we should try to remove it, especially now that we know that anything that does call it could be a performance problem.  But probably better to get this ticket in as-is, and then remove it and the option of passing `None` as stacklevel, since those may be breaking changes.

TallJimbo (approved):

python/lsst/utils/timer.py
timj: `_find_outside_stacklevel` won't be called by `timeMethod` any more but it could be used by anyone calling `logInfo` themselves if they don't provide the new `stacklevel` parameter.

python/lsst/utils/timer.py
timj: `stacklevel=1` would make the log message appear to come from `timeMethod_wrapper` but we want it to come from the thing that calls this since that would be the code in the file that added the `@timeMethod` decorator. I'll reword the comment. We do have tests for timeMethod and logInfo but I'll take a quick look at them.

PR #106, opened 2021-11-19
DM-32661: Optimization path should check values not keys

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-32661.perf.rst
+++ b/doc/changes/DM-32661.perf.rst
@@ -1,0 +1,1 @@
+Fixed an optimization when using `lsst.utils.TemplateMeta` classes with `isintance` or `issubclass`.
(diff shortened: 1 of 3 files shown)


timj:
@ktlim  is this what you had in mind? I'm not sure how to test it other than seeing if we now get coverage on the line that returns `True`.

timj:
@ktlim and indeed the coverage has got worse because now the optimization is working and the explicit loop never triggers. I don't know enough about this code to write a test that will now trigger the fallback code.

erykoff:
I don't see how the fallback code could ever trigger.  Isn't it just a loop over what the `in` is checking?

timj:
@erykoff, @TallJimbo gave me the clue about the fallback code and I've written some tests which do exercise it.

ktlim (approved):

PR #108, opened 2021-11-29
DM-32679: Call lsst.log traceSetAt from trace_set_at if available

timj:
This allows people to enable TRACE loggers in C++ code without having to know that such loggers exist and without people defensively calling logUtils traceSetAt as well as trace_set_at.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/logging.py
+++ b/python/lsst/utils/logging.py
@@ -26,6 +26,11 @@ from typing import (
     Union,
 )
 
+try:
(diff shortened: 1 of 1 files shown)


ktlim (approved):
codecov doesn't like you, but the code looks fine.

timj:
Codecov can never see lsst.log so I don't have any choice in the coverage and mocking lsst.log didn't seem worth it.

PR #109, opened 2021-12-03
DM-32801: Add sphinx docs build action

timj:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -1,0 +1,37 @@
+name: docs
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  build_sphinx_docs:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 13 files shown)


ktlim (approved):
I'm a little worried about all of the style/lint changes going on informally without a clear roadmap or policy, but this looks OK.

timj:
Most of the fixes here are fixing legitimate problems with the numpydoc so I'm not sure what the problem is.

ktlim:
pydocstyle has not been formally adopted (and barely even suggested, to my knowledge) as a tool for checking/verifying our docs.  I agree that it can help us get rid of a lot of warnings and other problems, but we still shouldn't add it in random places without a clear roadmap.

timj:
Right, but the usage here is consistent with the dev guide isn't it? What we haven't done is formalized the list of warnings that should be used -- the ones here and in daf_butler are adhoc.

PR #110, opened 2021-12-08
DM-32437: Reformat with black and isort and enable action to check

timj:
## Checklist

- [x] ran Jenkins


TallJimbo (approved):
I've made some small adjustments, and then managed to cherry-pick these commits (not the actual reformatting ones, of course) to the other packages.  Thanks!

PR #111, opened 2022-01-28
DM-33403: Move base.Packages to lsst.utils

timj:
This implementation can read old pickle files and force them to be frin the lsst.utils class.

## Checklist

- [x] ran Jenkins


python/lsst/utils/packages.py
```diff
@@ -0,0 +1,530 @@
...
+    Notes
+    -----
+    We wade through `sys.modules` and attempt to determine the version for each
+    module.  Note, therefore, that we can only report on modules that have
+    *already* been imported.
+
+    We don't include any module for which we cannot determine a version.
+    """
+    # Attempt to import libraries that only report their version in python
+    for module_name in PYTHON:
```
timj, line 110: So this is why galsim always turns up in the version information even when we aren't using it.

python/lsst/utils/packages.py
```diff
@@ -0,0 +1,538 @@
...
+
+    Notes
+    -----
+    We use EUPS to determine the version of certain products (those that don't
+    provide a means to determine the version any other way) and to check if
+    uninstalled packages are being used. We only report the product/version
+    for these packages.
+    """
+    try:
+        from eups import Eups
```
timj, line 171: The code coverage is poor mostly because in the GitHub test environment there aren't any EUPS packages set up. We can't set up `utils` itself because that would require `sconsUtils` (even if `cpputils` was switched to optional). Does it seem feasible to create a eups package and set it up just for the test environment?

ktlim: You can setup a local version of anything with a `ups/*.table` file.  Also, eups itself seems like it's locally setup.
```
$ mkdir -p test_package/ups
$ touch test_package/ups/test_package.table
$ cd test_package
$ setup -k -r .
$ eups list --setup
eups                  LOCAL:/home/ktlim/lsstsw2/miniconda/envs/lsst-scipipe-1.0.0/eups 	setup
test_package          LOCAL:/home/ktlim/lsstsw2/test_package 	setup
```
But I guess you'd actually need to clone the test package in order to test the git version code.

`lsst_build` has no dependencies and is designed to be cloned into an `lsstsw` directory, but I'm not sure you want to depend upon that existing.

tests/test_packages.py
```diff
@@ -0,0 +1,186 @@
...
+        # Loop over keys to check iterator.
+        keys = {k for k in new}
+        self.assertEqual(keys, set(dict(new).keys()))
+
+        # Loop over values to check that we do get them all.
+        values = {v for v in new.values()}
+        self.assertEqual(values, set(dict(new).values()))
+
+        # Serialize via bytes
+        for format in ("pickle", "yaml"):
```
ktlim, line 149: Test JSON too?

ktlim (approved):

python/lsst/utils/packages.py
```diff
@@ -0,0 +1,538 @@
...
+"""
+Determine which packages are being used in the system and their versions
+"""
+import hashlib
+import importlib
+import io
+import json
+import logging
+import os
+import pickle as pickle
```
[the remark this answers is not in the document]
timj: That was presumably a python2 to 3 leftover.

PR #112, opened 2022-02-09
DM-33331: Monitor memory usage at select steps of BPS submission

mxk62:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/timer.py
+++ b/python/lsst/utils/timer.py
@@ -15,7 +15,7 @@
 
 from __future__ import annotations
 
-__all__ = ["logInfo", "timeMethod", "time_this", "get_mem_usage"]
+__all__ = ["logInfo", "timeMethod", "time_this"]
 
 import datetime
 import functools
@@ -36,10 +36,15 @@ from typing import (
     Tuple,
 )
 
-import psutil
+from astropy import units as u
+
+from .usage import get_current_mem_usage, get_peak_mem_usage
 
 if TYPE_CHECKING:
     from .logging import LsstLoggers
+
+
+_LOG = logging.getLogger(__name__)
 
 
 def _add_to_metadata(metadata: MutableMapping, name: str, value: Any) -> None:
@@ -111,34 +116,6 @@ def _find_outside_stacklevel() -> int:
         del s
 
     return stacklevel
-
-
-def get_mem_usage(include_children: bool = False) -> int:
-    """Report current memory usage.
-
-    Parameters
-    ----------
-    include_children : `bool`, optional
-        Flag indicating if memory usage by child processes (if any) should be
-        included in the report. Be default, only memory usage of the calling
-        process is reported.
-
-    Returns
-    -------
-    mem : `int`
-        Memory usage expressed in bytes.
-
-    Notes
-    -----
-    Function reports current memory usage using resident set size as a proxy.
-    As such the value it reports is capped at available physical RAM and may
-    not reflect the actual memory allocated to the process.
-    """
-    proc = psutil.Process()
-    usage = proc.memory_info().rss
-    if include_children:
-        usage += sum([child.memory_info().rss for child in proc.children()])
-    return usage
 
 
 def logPairs(
@@ -395,7 +372,8 @@ def time_this(
     args: Iterable[Any] = (),
     mem_usage: bool = False,
     mem_child: bool = False,
-    mem_units: str = "B",
+    mem_unit: u.Quantity = u.byte,
+    mem_fmt: str = ".0f",
 ) -> Iterator[None]:
     """Time the enclosed block and issue a log message.
 
@@ -422,8 +400,11 @@ def time_this(
         Defaults, to False.
     mem_child : `bool`, optional
         Flag indication whether to include memory usage of the child processes.
-    mem_units : {"B", "KB", "MB", "GB"}, optional
-        Units to use when reporting the memory usage. Defaults to bytes ("B").
+    mem_unit : `astropy.units.Unit`, optional
+        Unit to use when reporting the memory usage. Defaults to bytes.
+    mem_fmt : `str`, optional
+        Format specifier to use when displaying values related to memory usage.
+        Defaults to '.0f'.
     """
     if log is None:
         log = logging.getLogger()
@@ -434,20 +415,24 @@ def time_this(
     success = False
     start = time.time()
     if mem_usage:
-        mem_start = get_mem_usage(include_children=mem_child)
+        current_usages_start = get_current_mem_usage()
+        peak_usages_start = get_peak_mem_usage()
+
     try:
         yield
         success = True
     finally:
         end = time.time()
-        if mem_usage:
-            mem_end = get_mem_usage(include_children=mem_child)
 
         # The message is pre-inserted to allow the logger to expand
         # the additional args provided. Make that easier by converting
         # the None message to empty string.
         if msg is None:
             msg = ""
+
+        # Convert user provided parameters (if any) to mutable sequence to make
+        # mypy stop complaining when additional parameters will be added below.
+        params = list(args) if args else []
 
         if not success:
             # Something went wrong so change the log level to indicate
@@ -456,26 +441,30 @@ def time_this(
 
         # Specify stacklevel to ensure the message is reported from the
         # caller (1 is this file, 2 is contextlib, 3 is user)
-        if mem_usage:
-            scales = {
-                "B": (1.0, "%d"),
-                "KB": (float(2**10), "%.1f"),
-                "MB": (float(2**20), "%.2f"),
-                "GB": (float(2**30), "%.3f"),
-            }
-            if mem_units not in scales:
-                mem_units = "B"
-            scale, fmt = scales[mem_units]
-            report = f"%sTook %.4f seconds; memory usage {fmt} {mem_units} (increment: {fmt} {mem_units})"
-            log.log(
-                level,
-                msg + report,
-                *args,
-                ": " if msg else "",
-                end - start,
-                mem_end / scale,
-                (mem_end - mem_start) / scale,
-                stacklevel=3,
+        params += (": " if msg else "", end - start)
+        msg += "%sTook %.4f seconds"
+        if mem_usage and log.isEnabledFor(level):
+            current_usages_end = get_current_mem_usage()
+            peak_usages_end = get_peak_mem_usage()
+
+            current_deltas = [end - start for end, start in zip(current_usages_end, current_usages_start)]
+            peak_deltas = [end - start for end, start in zip(peak_usages_end, peak_usages_start)]
+
+            current_usage = current_usages_end[0]
+            current_delta = current_deltas[0]
+            peak_delta = peak_deltas[0]
+            if mem_child:
+                current_usage += current_usages_end[1]
+                current_delta += current_deltas[1]
+                peak_delta += peak_deltas[1]
+
+            if not mem_unit.is_equivalent(u.byte):
+                _LOG.warning("Invalid memory unit '%s', using '%s' instead", mem_unit, u.byte)
+                mem_unit = u.byte
+
+            msg += (
+                f"; current memory usage: {current_usage.to(mem_unit):{mem_fmt}}"
+                f", delta: {current_delta.to(mem_unit):{mem_fmt}}"
+                f", peak delta: {peak_delta.to(mem_unit):{mem_fmt}}"
             )
-        else:
-            log.log(level, msg + "%sTook %.4f seconds", *args, ": " if msg else "", end - start, stacklevel=3)
+        log.log(level, msg, *params, stacklevel=3)
--- a/python/lsst/utils/timer.py
+++ b/python/lsst/utils/timer.py
@@ -15,7 +15,7 @@
 
 from __future__ import annotations
 
-__all__ = ["logInfo", "timeMethod", "time_this"]
+__all__ = ["logInfo", "timeMethod", "time_this", "get_mem_usage"]
 
 import datetime
 import functools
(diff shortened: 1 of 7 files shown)


ktlim:
Please fix the ticket number in the PR title.  (While the branch appears correct, I think some tools rely on the title.)

mxk62:
@ktlim Fixed.

python/lsst/utils/timer.py
```diff
@@ -412,4 +456,26 @@ def time_this(
 
         # Specify stacklevel to ensure the message is reported from the
         # caller (1 is this file, 2 is contextlib, 3 is user)
-        log.log(level, msg + "%sTook %.4f seconds", *args, ": " if msg else "", end - start, stacklevel=3)
+        if mem_usage:
+            scales = {
```
timj, line 460: `astropy.unit` would do away with all this. The downside is possibly additional overhead, although that would only trigger if we were logging and wanted the memory statistics.

python/lsst/utils/timer.py
```diff
@@ -412,4 +456,26 @@ def time_this(
 
         # Specify stacklevel to ensure the message is reported from the
         # caller (1 is this file, 2 is contextlib, 3 is user)
-        log.log(level, msg + "%sTook %.4f seconds", *args, ": " if msg else "", end - start, stacklevel=3)
+        if mem_usage:
```
timj, line 459: Because this has overhead (if we've moved the get_mem_usage down a bit), it would probably be worth doing  an `if log.isEnabledFor(level)` check in here -- we don't want the overhead if the log message is never going to appear

python/lsst/utils/timer.py
```diff
@@ -393,11 +433,15 @@ def time_this(
...
     start = time.time()
+    if mem_usage:
+        mem_start = get_mem_usage(include_children=mem_child)
     try:
         yield
         success = True
     finally:
         end = time.time()
+        if mem_usage:
+            mem_end = get_mem_usage(include_children=mem_child)
```
timj, line 444: Can we move this into the if block below and wrap it with `isEnabledFor` so we don't bother if the log message never appears. We can't add `isEnabledFor` here because the log level might change below.

python/lsst/utils/timer.py
```diff
@@ -412,4 +456,26 @@ def time_this(
...
+            scales = {
+                "B": (1.0, "%d"),
+                "KB": (float(2**10), "%.1f"),
+                "MB": (float(2**20), "%.2f"),
+                "GB": (float(2**30), "%.3f"),
+            }
+            if mem_units not in scales:
+                mem_units = "B"
+            scale, fmt = scales[mem_units]
+            report = f"%sTook %.4f seconds; memory usage {fmt} {mem_units} (increment: {fmt} {mem_units})"
```
timj, line 469: I think say "current memory usage" and then "delta" rather than "increment". Also, I'm a bit confused as to why this is an f-string (which should not be used for logging) but then below in the log call there are additional args like `(mem_end - mem_start)`.  What does the output look like? I'm sure I'm missing something.

I would prefer it if the code did something like:

```python
args += (": " if msg else "",)
msg += "%sTook %.4f seconds"
if mem_usage:
    if log.isEnabledFor(level):
        mem_end = ...
        msg += "; memory usage %s..."
        args += (end-start, mem_end / scale, ...)
log.log(level, msg, *args, stacklevel=3)
```
So we add to the args and message but then only call `log.log` once.

python/lsst/utils/timer.py
```diff
@@ -111,6 +113,34 @@ def _find_outside_stacklevel() -> int:
...
+    Parameters
+    ----------
+    include_children : `bool`, optional
+        Flag indicating if memory usage by child processes (if any) should be
+        included in the report. Be default, only memory usage of the calling
+        process is reported.
+
+    Returns
+    -------
+    mem : `int`
```
timj, line 128: Consider returning astropy Quantity to simplify reporting.

timj (approved):
Many of my comments are about reorganizing things as mentioned on the BPS review. I also think that the log output for time_this can be simplified.

One other thing that we might want to do is to remove the getrusage call out of `logInfo` and move that into the new `usage.py`. If you make it return a DataClass with all the entries that logPairs uses and then the get_peak_usage could either disappear or else call that and unpack the information from the DataClass.

python/lsst/utils/timer.py
```diff
@@ -111,6 +113,34 @@ def _find_outside_stacklevel() -> int:
...
+        Memory usage expressed in bytes.
+
+    Notes
+    -----
+    Function reports current memory usage using resident set size as a proxy.
+    As such the value it reports is capped at available physical RAM and may
+    not reflect the actual memory allocated to the process.
+    """
+    proc = psutil.Process()
```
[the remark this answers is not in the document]
mxk62: Comparing to the exection times of snippets I usually time with `time_this` it seems to be rather negligible:
```python
>>> import timeit
>>> timeit.timeit('psutil.Process()', setup='import psutil', number=1000)
0.01848633700137725
```

python/lsst/utils/timer.py
mxk62: Will fix the text according to your suggestions.

Regarding the use of the f-string. As I allowed the user to pick the units which will be used to report the memory usage I wanted to vary the accuracy of the numbers depending on said units. For example, using three decimal places makes sense for GB/GiB, but not much so for bytes. As I failed to find a way how to define format specifiers dynamically when using `%`-formatting I decided to use f-string to construct the log message up front. But to my understanding, I wasn't using the f-string directly in the log message itself. It was just receiving the regular string I constructed earlier with the desired specifiers. Or am I getting it wrong?

timj: I missed the `report` in the line below. The `isEnabledFor` is even more relevant in that case since there is no reason to form the f-string if it will not be used.

My main comment was whether astropy quantity would help -- it seems that quantities can accept format specifiers in f-strings:

```python
>>> import astropy.units as u
>>> mem = 5023553466234 * u.bytes
>>> print(f"{mem.to('GB'):.3f}")
5023.553 Gbyte
```

python/lsst/utils/timer.py
```diff
@@ -405,11 +426,42 @@ def time_this(
...
         if not success:
             # Something went wrong so change the log level to indicate
             # this.
             level = logging.ERROR
 
         # Specify stacklevel to ensure the message is reported from the
         # caller (1 is this file, 2 is contextlib, 3 is user)
-        log.log(level, msg + "%sTook %.4f seconds", *args, ": " if msg else "", end - start, stacklevel=3)
+        msg += "%sTook %.4f seconds; "
+        params += (": " if msg else "", end - start)
```
timj, line 441: Are you sure this logic is correct? Isn't `msg` by definition now set (the line before)? I think this line has to move before the previous line. Else where `if msg is None:` is instead do:

```
msg = msg + ": " if msg else ""
```
and remove the `if msg else` from this line (and the `%s` from the line before.

python/lsst/utils/timer.py
```diff
@@ -405,11 +426,42 @@ def time_this(
...
+        if mem_usage:
+            if log.isEnabledFor(level):
```
timj, line 443: maybe combine this with the previous `if` and reduce the indent level in the block below?

python/lsst/utils/usage.py
```diff
@@ -0,0 +1,72 @@
...
+    a proxy. As such the value it reports is capped at available physical RAM
+    and may not reflect the actual maximal value.
+    """
+    # Units getrusage(2) uses to report the maximum resident set size are
+    # platform dependent (kilobytes on Linux, bytes on OSX).
+    unit = u.kibibyte if platform.system() == "Linux" else u.byte
+
+    peak_main = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss * unit
+    peak_child = resource.getrusage(resource.RUSAGE_CHILDREN).ru_maxrss * unit
+    return peak_main.to(u.byte), peak_child.to(u.byte)
```
timj, line 72: Do we need to do this unit conversion? If you don't do it then everything should still work, or at least it's meant to. Attaching the units in whatever form they have and then deferring the conversion until it's needed is I think the better approach when we have units. Otherwise you are doing a conversion to bytes here and then a second conversion later.

python/lsst/utils/timer.py
```diff
@@ -405,11 +426,42 @@ def time_this(
...
+
         if not success:
             # Something went wrong so change the log level to indicate
             # this.
             level = logging.ERROR
 
         # Specify stacklevel to ensure the message is reported from the
         # caller (1 is this file, 2 is contextlib, 3 is user)
-        log.log(level, msg + "%sTook %.4f seconds", *args, ": " if msg else "", end - start, stacklevel=3)
+        msg += "%sTook %.4f seconds; "
```
timj, line 440: Does this result in `"; "` on the end of the log message even if memory reporting is not enabled? Can you move that to the block below and include it in the message that is added later?

python/lsst/utils/timer.py
```diff
@@ -405,11 +426,42 @@ def time_this(
...
+
+                current_usage = current_usages_end[0]
+                current_delta = current_deltas[0]
+                peak_delta = peak_deltas[0]
+                if mem_child:
+                    current_usage += current_usages_end[1]
+                    current_delta += current_deltas[1]
+                    peak_delta += peak_deltas[1]
+
+                if not mem_unit.is_equivalent(u.byte):
```
timj, line 458: If the caller is passing in a bad unit I'd be inclined to at least log a complaint? I understand that you don't want things to crash if the caller passes in a bad unit but how are they going to know they've passed in a bad unit?

python/lsst/utils/timer.py
```diff
@@ -405,11 +426,42 @@ def time_this(
...
+                    mem_unit = u.byte
+
+                msg += "current memory usage %s (delta: %s, peak delta: %s)"
```
timj, line 461: In this particular case, rather than formatting the strings here and then formatting the strings in the log message, we may as well do:

```python
msg += f"; current memory usage {current_usage.to(mem_unit):{mem_fmt}} ..."
```
since we know that the log message will be issued and doing one string substitution is presumably more efficient than doing two.

python/lsst/utils/timer.py
```diff
@@ -405,11 +426,42 @@ def time_this(
...
+                params += (
+                    f"{current_usage.to(mem_unit):{mem_fmt}}",
+                    f"{current_delta.to(mem_unit):{mem_fmt}}",
```
timj, line 464: In your testing is current_delta anything other than 0? I suppose it depends on what the garbage collector is doing.

tests/test_timer.py
```diff
@@ -188,6 +189,54 @@ def testTimer(self):
         self.assertIn("Took", cm.output[0])
         self.assertEqual(cm.records[0].filename, THIS_FILE)
 
+        # Report memory usage.
+        with self.assertLogs(level="DEBUG") as cm:
+            with time_this(level=logging.DEBUG, prefix=None, mem_usage=True):
+                pass
+        self.assertEqual(cm.records[0].name, "root")
+        self.assertEqual(cm.records[0].levelname, "DEBUG")
+        self.assertIn("Took", cm.output[0])
```
timj, line 198: and for at least one test do `self.assertNotIn(": Took", cm.output[0])` when no message is given (so that we don't get the `:` by mistake. Solely because I'm worried that the code currently always includes it. Similarly for the `"; "`.

timj (approved):
Looks great. Some minor comments.

python/lsst/utils/timer.py
```diff
@@ -405,11 +426,42 @@ def time_this(
...
+                current_deltas = [end - start for end, start in zip(current_usages_end, current_usages_start)]
+                peak_deltas = [end - start for end, start in zip(peak_usages_end, peak_usages_start)]
+
+                current_usage = current_usages_end[0]
+                current_delta = current_deltas[0]
+                peak_delta = peak_deltas[0]
+                if mem_child:
+                    current_usage += current_usages_end[1]
+                    current_delta += current_deltas[1]
+                    peak_delta += peak_deltas[1]
```
timj, line 456: Presumably this is not necessarily the maximum memory needed by all the children when running simultaneously? It's the sum of the peak of the current process and the largest peak for any of the children.

python/lsst/utils/usage.py
mxk62: You're right, enforcing conversion to bytes is not really necessary. Will fix.

python/lsst/utils/timer.py
mxk62: You're right. Will fix.

python/lsst/utils/timer.py
mxk62: Yes, it's the sum of the peak of the current process and the largest peak for any of the children.

python/lsst/utils/timer.py
mxk62: Yes, it can be larger then 0. Here's an example log from a test BPS submission:
```
...
lsst.ctrl.bps.drivers INFO: Initializing execution environment completed: Took 0.2665 seconds; current memory usage: 0.231 Gibyte, delta: 0.023 Gibyte, peak delta: 0.023 Gibyte
...
lsst.ctrl.bps.drivers INFO: Acquire stage completed: Took 5.5747 seconds; current memory usage: 0.365 Gibyte, delta: 0.134 Gibyte, peak delta: 0.134 Gibyte
...
```
You're right that for "isolated" snippets, `current_delta` mostly likely reflects what the garbage collector is doing. But it also can give you some (crude) idea how memory usage changes during execution of your code. For example, the log above shows that getting the quantum graph into the memory (that's what acquire stage in `ctrl_bps` does) took approx. 100 MiB.

PR #115, opened 2022-02-25
DM-20970: Fix units in max_rss field for logInfo

timj:
This moves the `getrusage` code from timer.py into usage.py and ensures that the units are in bytes.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/usage.py
+++ b/python/lsst/utils/usage.py
@@ -14,9 +14,11 @@
 
 __all__ = ["get_current_mem_usage", "get_peak_mem_usage"]
 
+import dataclasses
 import platform
 import resource
-from typing import Tuple
+import time
+from typing import Dict, Tuple, Union
 
 import astropy.units as u
 import psutil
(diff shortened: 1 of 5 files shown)


python/lsst/utils/usage.py
```diff
@@ -61,10 +88,60 @@ def get_peak_mem_usage() -> Tuple[u.Quantity, u.Quantity]:
...
     and may not reflect the actual maximal value.
     """
-    # Units getrusage(2) uses to report the maximum resident set size are
-    # platform dependent (kilobytes on Linux, bytes on OSX).
-    unit = u.kibibyte if platform.system() == "Linux" else u.byte
-
-    peak_main = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss * unit
-    peak_child = resource.getrusage(resource.RUSAGE_CHILDREN).ru_maxrss * unit
+    peak_main = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss * _RUSAGE_MEMORY_UNIT
+    peak_child = resource.getrusage(resource.RUSAGE_CHILDREN).ru_maxrss * _RUSAGE_MEMORY_UNIT
```
timj, line 92: I'm torn as to whether these two lines should call `_get_current_rusage`. It presumably has a small additional overhead and means we apply the units twice (once to convert from machine specific to bytes and then applying the bytes again to turn this into a quantity).

python/lsst/utils/usage.py
```diff
@@ -12,14 +12,40 @@
...
-from astropy import units as u
 
-__all__ = ["get_current_mem_usage", "get_peak_mem_usage"]
+
+def _get_rusage_unit() -> u.Unit:
+    """Return the unit to use for memory usage returned by getrusage.
+
+    Returns
+    -------
+    unit : `astropy.units.Uni`
```
ktlim, line 32: Missing a "t"?

While the astropy units framework is clear and error-reducing, I'm a tiny bit concerned about astropy being sucked in at this very low level of the stack, and for something that is essentially picking a platform-dependent constant multiplier.

python/lsst/utils/usage.py
```diff
@@ -61,10 +88,60 @@ def get_peak_mem_usage() -> Tuple[u.Quantity, u.Quantity]:
...
+    peak_main = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss * _RUSAGE_MEMORY_UNIT
+    peak_child = resource.getrusage(resource.RUSAGE_CHILDREN).ru_maxrss * _RUSAGE_MEMORY_UNIT
     return peak_main, peak_child
+
+
+@dataclasses.dataclass(frozen=True)
+class _UsageInfo:
+    """Summary of process usage."""
+
+    cpuTime: float
```
ktlim, line 100: I guess it's obvious that these are in seconds?

ktlim (approved):

python/lsst/utils/usage.py
timj: I've pondered this over the weekend and I agree that using Quantity in this bit of code is not going to help much. I'll remove it.

PR #116, opened 2022-03-03
DM-33919: Add helper class for periodic logging

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/logging.py
+++ b/python/lsst/utils/logging.py
@@ -11,9 +11,10 @@
 
 from __future__ import annotations
 
-__all__ = ("TRACE", "VERBOSE", "getLogger", "LsstLogAdapter", "trace_set_at")
+__all__ = ("TRACE", "VERBOSE", "getLogger", "LsstLogAdapter", "PeriodicLogger", "trace_set_at")
 
 import logging
+import time
 from contextlib import contextmanager
 from logging import LoggerAdapter
 from typing import Any, Generator, List, Optional, Union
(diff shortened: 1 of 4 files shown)


python/lsst/utils/logging.py
```diff
@@ -332,3 +333,57 @@ def getLogger(name: Optional[str] = None, logger: logging.Logger = None) -> Lsst
 
 
 LsstLoggers = Union[logging.Logger, LsstLogAdapter]
+
+
+class PeriodicLogging:
```
arunkannawadi, line 338: Should this be `PeriodicLogger` may be?

python/lsst/utils/logging.py
```diff
@@ -332,3 +333,57 @@ def getLogger(name: Optional[str] = None, logger: logging.Logger = None) -> Lsst
...
+        Logger to use when issuing a message.
+    interval : `float`
+        The minimum interval between log messages. If `None` the class
+        default will be used.
+    level : `int`, optional
+        Log level to use when issuing messages.
+    """
+
+    LOGGING_INTERVAL = 600.0
+    """Default interval between log messages."""
```
arunkannawadi, line 357: The dev guide says that the comment (in this case docstring) should be about the code that follows it.

arunkannawadi (approved):
This is a much nicer solution to log heartbeat messages!

python/lsst/utils/logging.py
timj: I'm fairly sure that Sphinx docstrings are meant to follow the thing they are referring to: https://developer.lsst.io/python/numpydoc.html?highlight=sphinx#placement-of-constant-and-class-attribute-docstrings

I checked the sphinx output and this text does appear with `LOGGING_INTERVAL`.

python/lsst/utils/logging.py
timj: Maybe. Although it's not a subclass of `logging.Logger`... I'll change it.

python/lsst/utils/logging.py
arunkannawadi: Perhaps that was only for comments then.

timj: Yes. Comments come above but docstrings come after. It's all very odd.

PR #118, opened 2022-03-07
DM-33950: Ensure that periodic logger reports the caller filename

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/logging.py
+++ b/python/lsst/utils/logging.py
@@ -363,6 +363,11 @@ class PeriodicLogger:
         self.next_log_time = time.time() + self.interval
         self.num_issued = 0
 
+        # The stacklevel we need to issue logs is determined by the type
+        # of logger we have been given. A LoggerAdapter has an extra
+        # level of indirection.
+        self._stacklevel = 3 if isinstance(self.logger, LoggerAdapter) else 2
+
     def log(self, msg: str, *args: Any) -> bool:
         """Issue a log message if the interval has elapsed.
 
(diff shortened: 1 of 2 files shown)


python/lsst/utils/logging.py
```diff
@@ -363,6 +363,11 @@ def __init__(self, logger: LsstLoggers, interval: Optional[float] = None, level:
         self.next_log_time = time.time() + self.interval
         self.num_issued = 0
 
+        # The stacklevel we need to issue logs is determined by the type
+        # of logger we have been given. A LoggerAdapter has an extra
+        # level of indirection.
+        self._stacklevel = 3 if isinstance(self.logger, LoggerAdapter) else 2
```
arunkannawadi, line 369: I wouldn't know where these numbers came from. I know that it has to be one level higher up, and wonder if there's way to specify that directly instead of the stack levels themselves.

timj: The default is `stacklevel=1` which is `Logger.log()` reporting the stack information from its caller. We are calling `a_logger.log()` and so we don't want the message to look like it's coming from that line we want it from our caller and so need to use `stacklevel=2`. A `logging.LoggerAdapter` has an extra call between us an `Logger.log()` (it's arguably a bug that `LoggerAdapter` does not correct for this) so we need to put in a 3. The caller is not meant to know any of this and we waren't expecting the caller to pass in kwargs and override the stacklevel. You'll see lots of other places in this file where `stacklevel` is set to ensure that the log message looks like it is coming from the correct place.

I'm not really following what you are asking.

arunkannawadi: My comment was that the PeriodicLogger ideally subtracts 1 from the "current" stacklevel instead of setting it to 3 or 2 as the case may be. I don't know logging well enough to know if that's possible though.

arunkannawadi (approved):

python/lsst/utils/logging.py
arunkannawadi: Never mind, I had misunderstood what the `stacklevel` argument stood for. I looked at the docs for logging and this _is_ exactly what I was asking for. Sorry for the noise.

PR #123, opened 2022-05-25
DM-12659: Remove C++ support from pure-Python package.

kfindeisen:
This PR removes vestigial support for `utils` include and library files from the `ups/` directory, bringing it in line with the template pure-Python package.


timj (approved):
Looks great.

timj (approved):

PR #126, opened 2022-06-10
Add upload to pypi

mwittgen:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - "*"
   pull_request:
 
 jobs:
@@ -80,3 +82,37 @@ jobs:
         uses: codecov/codecov-action@v2
         with:
           file: ./coverage.xml
+
+  pypi:
+
+    runs-on: ubuntu-latest
+    needs: [build_and_test]
+    if: startsWith(github.ref, 'refs/tags/')
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # Need to clone everything to embed the version.
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v3
+        with:
+          python-version: 3.8
+          cache: "pip"
+          cache-dependency-path: "setup.cfg"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install --upgrade setuptools wheel build
+
+      - name: Build and create distribution
+        run: |
+          python -m build --sdist --wheel
+
+      - name: Upload
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_UPLOADS }}
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,7 @@
-See COPYRIGHT file for code ownership.
+BSD 3-Clause License
+
+Copyright (c) 2022, lsst
+All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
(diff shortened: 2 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -80,3 +82,37 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # Need to clone everything to embed the version.
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v3
+        with:
+          python-version: 3.8
```
timj, line 101: Does this python version matter for anything? Should it match build_and_test or is it irrelevant because the code in the package is never run?

LICENSE
```diff
@@ -1,26 +1,27 @@
-See COPYRIGHT file for code ownership.
+For copyright information see the COPYRIGHT file included in the top-level
+directory of this distribution.
 
-Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions are met:
+Redistribution and use in source and binary forms, with or without modification,
```
timj, line 4: What was wrong with the formatting of the LICENSE file?

setup.cfg
```diff
@@ -1,14 +1,27 @@
...
+    Source code =   https://github.com/lsst/lsst_versions
+    Issue tracker = https://github.com/lsst/lsst_versions/issues
 classifiers =
-    License :: OSI Approved ::  GNU General Public License v3 or later (GPLv3+)
+    Intended Audience :: Developers
+    License :: OSI Approved :: BSD License
     Operating System :: OS Independent
     Programming Language :: Python :: 3
     Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
```
timj, line 17: We probably need to start using tox and testing multiple python versions in a matrix (or at least not using tox but still using a matrix). I'm fine with you doing that on a later ticket.

setup.cfg
```diff
@@ -1,14 +1,27 @@
 [metadata]
-name = lsst_utils
+name = lsst-utils
 description = Utility functions from Rubin Observatory Data Management for the Legacy Survey of Space and Time (LSST).
 author = Rubin Observatory Data Management
+author_email = dm-admin@lists.lsst.org
 url = https://github.com/lsst/utils
+license = BSD 3-Clause License
+project_urls =
+    Source code =   https://github.com/lsst/lsst_versions
```
timj, line 9: These are wrong.

timj (approved):
Looks okay. One actual bug in the cfg file plus some questions.

.github/workflows/build.yaml
mwittgen: No. We can use python 3.10 here

LICENSE
mwittgen: I wanted to adapt the same LICENSE file. Maybe this should be reverted back.

.github/workflows/build.yaml
mwittgen: Yes. I would make this an additional ticket.

LICENSE
timj: I have no real idea how I ended up with a license version that is formatted differently to others. Can you make it a different commit though since it's not related to the pypi changes in the same commit?

mwittgen: Yes.

setup.cfg
mwittgen: I don't think we should define those.
The issue tracker link makes no sense
and `url` is sufficient to point to github

timj: Okay. Remove them then. Not sure why I added them in lsst-versions. Seems like PyPI is clever enough to understand a GitHub URL regardless.

LICENSE
mwittgen: Reverted

timj: I didn't say revert, I said put on a different commit on this branch.

LICENSE
```diff
@@ -1,4 +1,7 @@
-See COPYRIGHT file for code ownership.
+BSD 3-Clause License
+
+Copyright (c) 2022, lsst
+All rights reserved.
```
mwittgen, line 4: Much nicer with the license version explicitly stated.

PR #131, opened 2022-08-04
DM-35697: Relocate profile context manager from pipe_base

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/timer.py
+++ b/python/lsst/utils/timer.py
@@ -472,14 +472,16 @@ def time_this(
 
 
 @contextmanager
-def profile(filename: str, log: Optional[logging.Logger] = None) -> Iterator[Optional[cProfile.Profile]]:
-    """Profile the enclosed code block.
+def profile(
+    filename: Optional[str], log: Optional[logging.Logger] = None
+) -> Iterator[Optional[cProfile.Profile]]:
+    """Profile the enclosed code block and save the result to a file.
 
     Parameters
     ----------
-    filename : `str`
+    filename : `str` or `None`
         Filename to which to write profile (profiling disabled if `None` or
-        empty).
+        empty string).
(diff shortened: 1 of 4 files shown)


python/lsst/utils/timer.py
```diff
@@ -467,3 +469,55 @@ def time_this(
...
+    .. code-block:: bash
+
+        python -c 'import pstats; \
+            pstats.Stats("<filename>").sort_stats("cumtime").print_stats(30)'
+    """
+    if not filename:
+        # Nothing to do
+        yield None
+        return
+    from cProfile import Profile
```
parejkoj, line 513: Is there a reason not to have this import at the top?

tests/test_timer.py
```diff
@@ -287,5 +288,22 @@ def testTimer(self):
...
+            pass
+        self.assertIsNone(prof)
+
+        with tempfile.NamedTemporaryFile() as tmp:
+            with self.assertLogs("profile", level=logging.INFO) as cm:
+                with profile(tmp.name, logger) as prof:
+                    pass
+            self.assertEqual(len(cm.output), 2)
+            self.assertIsNotNone(prof)
+            self.assertTrue(os.path.exists(tmp.name))
```
parejkoj, line 305: I don't know if it's worth it, but maybe `self.assertIsInstance(pstats.Stats(tmp.name), pstats.Stats)`, just to confirm that an actual profiling file got written?

parejkoj (commented):
A few comments on the python, even though I think this was just moved over from elsewhere.

"ping" commit should be "pin"?

python/lsst/utils/timer.py
timj: I assume @PaulPrice did not want to have the overhead of importing another python module that is very unlikely to be used.

PR #137, opened 2022-09-05
DM-36108: move Ellipsis typing workaround from daf_butler

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-36108.misc.md
+++ b/doc/changes/DM-36108.misc.md
@@ -1,0 +1,1 @@
+Moved a module with a typing workaround for the built-in `Ellipsis` (`...`) singleton here, from `daf_butler`.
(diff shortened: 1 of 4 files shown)


python/lsst/utils/sets/__init__.py
```diff
@@ -0,0 +1,10 @@
+# This file is part of utils.
```
timj, line 1: Did this get committed by mistake from some other upcoming change?

timj (approved):
Looks okay but can you have a test that imports the code so that we know it will import?

python/lsst/utils/sets/__init__.py
TallJimbo: It's from a container class I later decided not to add at all; glad you caught it.

PR #138, opened 2022-09-06
DM-36116: Add docs for ellipsis workaround.

TallJimbo:
This replaces an outdated code comment with a module docstring, since it doesn't look possible to actually document the symbols exported by this module.

## Checklist

- [ ] ran Jenkins
- [ ] ~added a release note for user-visible changes to `doc/changes`~: I don't think it's user-visibly distinct from DM-36108.

### DIFF
--- a/python/lsst/utils/ellipsis.py
+++ b/python/lsst/utils/ellipsis.py
@@ -9,6 +9,28 @@
 # Use of this source code is governed by a 3-clause BSD-style
 # license that can be found in the LICENSE file.
 
+"""
+A type-annotation workaround for ``...`` not having an exposed type in Python.
+
+This module provides ``Ellipsis`` and ``EllipsisType`` symbols that are
+conditionally defined to point to the built-in "``...``" singleton and its type
+(respectively) at runtime, and an enum class and instance if
+`typing.TYPE_CHECKING` is `True` (an approach first suggested
+`here <https://github.com/python/typing/issues/684#issuecomment-548203158>`_).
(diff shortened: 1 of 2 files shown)


python/lsst/utils/ellipsis.py
```diff
@@ -9,23 +9,36 @@
...
+This module provides ``Ellipsis`` and ``EllipsisType`` symbols that are
+conditionally defined to point to the built-in "``...``" singleton and its type
+(respectively) at runtime, and an enum class and instance if
+`typing.TYPE_CHECKING` is `True` (an approach first suggested
+`here <https://github.com/python/typing/issues/684#issuecomment-548203158>`_).
+Type checkers should recognize enum literals as singletons, making this pair
+behave as expected under ``is`` comparisons and type-narrowing expressions,
+such as::
+
+    v: EllipseType | int
```
timj, line 24: Typo here. EllipseType.

python/lsst/utils/ellipsis.py
```diff
@@ -9,23 +9,36 @@
...
+    if v is not Ellipse:
```
timj, line 25: Ellipse?

timj (approved):
These changes look good but now I realize that `types.EllipsisType` does existing python 3.10 -- shouldn't we be using it here if it is defined and only falling back to the existing code if it's an older python?

https://docs.python.org/3/library/types.html#types.EllipsisType and https://docs.python.org/3/library/constants.html#Ellipsis

PR #145, opened 2023-02-13
DM-37987: Update to support python 3.11

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/logging.py
+++ b/python/lsst/utils/logging.py
@@ -33,14 +33,6 @@ try:
 except ImportError:
     logUtils = None
 
-# Python 3.11 understands internal stack frames in the logging.py
-# file and so does not need special offsetting of the stack level
-# when a LoggerAdapter is used.
-if sys.version_info < (3, 11, 0):
-    _OFFSET_STACK = True
-else:
-    _OFFSET_STACK = False
-
 
 # log level for trace (verbose debug).
 TRACE = 5
@@ -79,7 +71,7 @@ def _calculate_base_stacklevel(default: int, offset: int) -> int:
     to understand internal implementation details.
     """
     stacklevel = default
-    if _OFFSET_STACK:
+    if sys.version_info < (3, 11, 0):
         stacklevel += offset
     return stacklevel
 
--- a/python/lsst/utils/logging.py
+++ b/python/lsst/utils/logging.py
@@ -22,6 +22,7 @@ __all__ = (
 )
 
 import logging
+import sys
 import time
 from contextlib import contextmanager
 from logging import LoggerAdapter
@@ -31,6 +32,15 @@ try:
     import lsst.log.utils as logUtils
 except ImportError:
     logUtils = None
+
+# Python 3.11 understands internal stack frames in the logging.py
+# file and so does not need special offsetting of the stack level
+# when a LoggerAdapter is used.
+if sys.version_info < (3, 11, 0):
+    _OFFSET_STACK = True
+else:
+    _OFFSET_STACK = False
+
 
 # log level for trace (verbose debug).
 TRACE = 5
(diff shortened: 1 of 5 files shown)


python/lsst/utils/logging.py
```diff
@@ -320,8 +345,12 @@ def __init__(self, logger: LsstLoggers, interval: Optional[float] = None, level:
 
         # The stacklevel we need to issue logs is determined by the type
         # of logger we have been given. A LoggerAdapter has an extra
-        # level of indirection.
-        self._stacklevel = 3 if isinstance(self.logger, LoggerAdapter) else 2
+        # level of indirection. In Python 3.11 the logging infrastructure
+        # takes care to check for internal logging stack frames so there
+        # is no need for a difference.
+        stack_level = 2
```
ktlim, line 351: This is the same calculation as is used to generate the default level in `LsstLoggerAdapter` (and is thus the same as returned by `LsstLoggerAdapter._process_stacklevel()` with no argument).  I think it would be better if it is unified somehow rather than duplicated.

python/lsst/utils/logging.py
```diff
@@ -166,12 +182,22 @@ def getChild(self, name: str) -> LsstLogAdapter:
...
         # Python does not provide this method in LoggerAdapter but does
-        # not formally deprecated it in favor of critical() either.
+        # not formally deprecate it in favor of critical() either.
         # Provide it without deprecation message for consistency with Python.
-        # stacklevel=5 accounts for the forwarding of LoggerAdapter.
-        self.critical(msg, *args, **kwargs, stacklevel=4)
+        # Have to adjust stacklevel on Python 3.10 and older to account
+        # for call through self.critical.
+        stacklevel = self._process_stacklevel(kwargs)
+        stacklevel = stacklevel + 1 if _OFFSET_STACK else stacklevel
```
ktlim, line 199: I'd kind of like to see this `1` be an argument to `_process_stacklevel()` so that all `_OFFSET_STACK` handling can be limited to that function.

python/lsst/utils/logging.py
```diff
@@ -166,12 +182,22 @@ def getChild(self, name: str) -> LsstLogAdapter:
         """
         return getLogger(name=name, logger=self.logger)
 
+    def _process_stacklevel(self, kwargs: dict[str, Any]) -> int:
+        # Return default stacklevel, taking into account kwargs[stacklevel].
+        stacklevel = self._stacklevel
+        if "stacklevel" in kwargs:
+            stacklevel = stacklevel + kwargs.pop("stacklevel") - 1
```
ktlim, line 189: It's not clear where the `- 1` comes from here.  Can you add a comment to clarify?

ktlim (changes requested):

python/lsst/utils/logging.py
```diff
@@ -166,12 +215,30 @@ def getChild(self, name: str) -> LsstLogAdapter:
...
+        if "stacklevel" in kwargs:
+            # External user expects stacklevel=1 to mean "report from their
+            # line" but the code here is already trying to achieve that by
+            # default. Therefore if an external stacklevel is specified we
+            # adjust their stacklevel request by 1.
+            stacklevel = stacklevel + kwargs.pop("stacklevel") - 1
+
+        # The offset can be used to indicate that we have to take into account
+        # additional internal layers before calling python logging.
+        if _OFFSET_STACK:
```
ktlim, line 230: Not sure it helps much, but this could be converted to a call to `_calculate_base_stacklevel(stacklevel, offset)`, so there's only one use of `_OFFSET_STACK`, no?

ktlim (approved):

python/lsst/utils/logging.py
timj: Yes, I should have spotted that.

PR #151, opened 2023-03-29
DM-38386: Add utility function to calculate safe plotting limits

mfisherlevine:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/plotting/limits.py
+++ b/python/lsst/utils/plotting/limits.py
@@ -52,7 +52,7 @@ def calculate_safe_plotting_limits(
         The value to set the ylim maximum to.
     """
     if not isinstance(data_series, Iterable):
-        raise ValueError("data_series must be either an iterable, or an iterable of iterables")
+        raise TypeError("data_series must be either an iterable, or an iterable of iterables")
 
     # now we're sure we have an iterable, if it's just one make it a list of it
     # lsst.utils.ensure_iterable is not suitable here as we already have one,
--- a/python/lsst/utils/plotting/limits.py
+++ b/python/lsst/utils/plotting/limits.py
@@ -32,7 +32,7 @@ def calculate_safe_plotting_limits(
 
     Parameters
     ----------
-    data_series : `iterable` or `iterable` or `iterable`
+    data_series : `iterable` or `iterable` of `iterable`
         One or more data series which will be going on the same axis, and
         therefore want to have their common plotting limits calculated.
     percentile : `float`, optional
--- a/python/lsst/utils/plotting/limits.py
+++ b/python/lsst/utils/plotting/limits.py
@@ -12,7 +12,7 @@
 from __future__ import annotations
 
 from collections.abc import Sequence
-from typing import Optional
+from typing import Iterable, Optional
 
 import numpy as np
 
@@ -51,14 +51,16 @@ def calculate_safe_plotting_limits(
     ymax : `float`
         The value to set the ylim maximum to.
     """
-    if not isinstance(data_series, Sequence):
+    if not isinstance(data_series, Iterable):
         raise ValueError("data_series must be either an iterable, or an iterable of iterables")
 
     # now we're sure we have an iterable, if it's just one make it a list of it
     # lsst.utils.ensure_iterable is not suitable here as we already have one,
     # we would need ensure_iterable_of_iterables here
-    if not isinstance(data_series[0], Sequence):  # we have a single data series, not multiple
-        data_series = [data_series]  # now we can iterate like we had more than one
+    if not isinstance(data_series[0], Iterable):  # np.array are Iterable but not Sequence so isinstance that
+        # we have a single data series, not multiple, wrap in [] so we can
+        # iterate over it as if we were given many
+        data_series = [data_series]  #
 
     mins = []
     maxs = []
--- a/python/lsst/utils/plotting/limits.py
+++ b/python/lsst/utils/plotting/limits.py
@@ -1,0 +1,87 @@
+# This file is part of utils.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# Use of this source code is governed by a 3-clause BSD-style
+# license that can be found in the LICENSE file.
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+from typing import Optional
+
+import numpy as np
+
+
+def calculate_safe_plotting_limits(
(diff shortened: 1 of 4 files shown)


python/lsst/utils/plotting/limits.py
```diff
@@ -0,0 +1,79 @@
...
+#
+# Use of this source code is governed by a 3-clause BSD-style
+# license that can be found in the LICENSE file.
+
+from typing import Iterable
+
+import numpy as np
+
+
+def calculateSafePlottingLimits(dataSeries, percentile=99.9, constantExtra=None, symmetricAroundZero=False):
```
timj, line 17:
```suggestion
def calculateSafePlottingLimits(dataSeries, percentile=99.9, constantExtra=None, symmetricAroundZero=False) -> tuple[float, float]:
```

timj: You may also need `from __future__ import annotations` at the top since we still use python 3.8.

timj: Also will need to import `Optional` from `typing` up top. I'm not really sure what `dataSeries` is an iterable of. Floats? So maybe `Union[Iterable[float], Iterable[Iterable[float]]` ?

timj:
This PR is going to need a test file. I note that this is not a plotting function at all but a statistics function, so is plotting the right namespace for it? Also this is a new function in utils so should use snake case.

python/lsst/utils/plotting/limits.py
TallJimbo: Instead of `Optional`, use `float | None` and avoid that import.  And if `dataSeries` is effectively "numpy array-like", then `typing.Any` may be our best bet.  Others may have had better (and more extensive) experiences, but I'm pretty skeptical of the value of typing anything numpy-related.

timj: except we still support python 3.8 and I don't know if that works there.

tests/test_plotting_limits.py
```diff
@@ -0,0 +1,76 @@
...
+        self.assertAlmostEqual(strictMax, self.series1_max, places=4)
+
+        for extra in [-2.123, -1, 0, 1, 1.5, 23]:
+            ymin, ymax = calculate_safe_plotting_limits([self.series1, self.outliers], constant_extra=extra)
+            self.assertAlmostEqual(ymin, self.series1_min - extra, places=4)
+            self.assertAlmostEqual(ymax, self.series1_max + extra, places=4)
+
+
+if __name__ == "__main__":
+    unittest.main()
```
timj, line 76: I've been meaning to do this but you can fix this coverage warning by adding this to the end of `pyproject.toml`:

```toml
[tool.coverage.report]
show_missing = true
exclude_lines = [
    "pragma: no cover",
    "raise AssertionError",
    "raise NotImplementedError",
    "if __name__ == .__main__.:",
    "if TYPE_CHECKING:",
]
```

mfisherlevine: Done, thanks.

python/lsst/utils/plotting/limits.py
```diff
@@ -0,0 +1,89 @@
...
+    ymax : `float`
+        The value to set the ylim maximum to.
+    """
+    if not isinstance(data_series, Iterable):
+        raise ValueError("data_series must be either an iterable, or an iterable of iterables")
+
+    # now we're sure we have an iterable, if it's just one make it a list of it
+    # lsst.utils.ensure_iterable is not suitable here as we already have one,
+    # we would need ensure_iterable_of_iterables here
+    if not isinstance(data_series[0], Iterable):  # np.array are Iterable but not Sequence so isinstance that
```
timj, line 60: I am concerned that we are being a bit inconsistent. Remind me what the type error was if Iterable is used? numpy not being a Sequence is a bit odd.

python/lsst/utils/plotting/limits.py
```diff
@@ -0,0 +1,89 @@
...
+        # we have a single data series, not multiple, wrap in [] so we can
+        # iterate over it as if we were given many
+        data_series = [data_series]  #
```
timj, line 63: What's the purpose of that trailing `#` ?

timj (approved):
Looks good.

python/lsst/utils/plotting/limits.py
```diff
@@ -0,0 +1,89 @@
...
+        Make the limits symmetric around zero?
+    Returns
+    -------
+    ymin : `float`
+        The value to set the ylim minimum to.
+    ymax : `float`
+        The value to set the ylim maximum to.
+    """
+    if not isinstance(data_series, Iterable):
+        raise ValueError("data_series must be either an iterable, or an iterable of iterables")
```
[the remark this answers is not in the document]
mfisherlevine: Done, and changed it to a `TypeError` too, because, well, it is 🙂

python/lsst/utils/plotting/limits.py
mfisherlevine: Yes, I hate hate hate this. Getting this to work with a list, a numpy array, and a column from a pandas.dataframe, and an iterable of all of these was not super fun. This was the best solution I found. It is quite likely someone who knows what they're doing with mypy could do better, but I could not.

python/lsst/utils/plotting/limits.py
mfisherlevine: lol, I moved a comment and forgot that. It's gone now.

python/lsst/utils/plotting/limits.py
```diff
@@ -0,0 +1,89 @@
...
+        # we have a single data series, not multiple, wrap in [] so we can
+        # iterate over it as if we were given many
+        data_series = [data_series]  #
+
+    mins = []
+    maxs = []
+
+    for data in data_series:
+        max_val = np.nanpercentile(data, percentile)
+        min_val = np.nanpercentile(data, 100 - percentile)
```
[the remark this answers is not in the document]
mfisherlevine: Sure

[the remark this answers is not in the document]
mfisherlevine: Except no, it has to be `np.nanpercentile(data, 100.0 - percentile)` - `black` says so 🙃

python/lsst/utils/plotting/limits.py
TallJimbo: I am absolutely not suggesting a change to this particular code (especially now that it's working and merged), but in general when you run into trouble making a whole host of differently-behaving argument types work in single function, it might be time to make it into multiple functions that share code.  Python was not designed to support function overloading, and while we often want it to avoid having to come up with or export multiple names, sometimes that's the better price to pay.

PR #153, opened 2023-04-12
DM-38665: No longer use assert for python version mismatch

timj:
We can't actually guarantee that the ._version matches the parent version (it doesn't for pooch) so instead assume that the parent version is always right if there is a real mismatch.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/packages.py
+++ b/python/lsst/utils/packages.py
@@ -127,19 +127,31 @@ def getPythonPackages() -> Dict[str, str]:
         # Remove "foo.bar.version" in favor of "foo.bar"
         # This prevents duplication when the __init__.py includes
         # "from .version import *"
+        modified = False
         for ending in (".version", "._version"):
             if name.endswith(ending):
                 name = name[: -len(ending)]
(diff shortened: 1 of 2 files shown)


ktlim (approved):

python/lsst/utils/packages.py
```diff
@@ -127,19 +127,31 @@ def getPythonPackages() -> Dict[str, str]:
         # Remove "foo.bar.version" in favor of "foo.bar"
         # This prevents duplication when the __init__.py includes
         # "from .version import *"
+        modified = False
         for ending in (".version", "._version"):
             if name.endswith(ending):
                 name = name[: -len(ending)]
```
ktlim, line 133: Is `removesuffix()` more appropriate?

timj: Yes, although it was only added in python 3.9 and technically `utils` still supports python 3.8 -- I'm happy to drop old versions if it makes things easier but this seems like it doesn't make a huge gain in clarity.

PR #154, opened 2023-04-19
DM-38764: Add exclude list to MemoryTestCase.

ktlim:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/tests.py
+++ b/python/lsst/utils/tests.py
@@ -32,6 +32,7 @@ import gc
 import inspect
 import itertools
 import os
+import re
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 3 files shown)


python/lsst/utils/tests.py
```diff
@@ -121,13 +122,21 @@ def suiteClassWrapper(tests):
 class MemoryTestCase(unittest.TestCase):
     """Check for resource leaks."""
 
+    # List of regexps to ignore when checking for open files.
```
timj, line 125: Should this be a triple-quoted string on the line after the `ignore_regexps` definition?

python/lsst/utils/tests.py
```diff
@@ -121,13 +122,21 @@ def suiteClassWrapper(tests):
...
+    ignore_regexps = []
```
timj, line 126:
```suggestion
    ignore_regexps: list[str] = []
```
? To make it a bit more explicit when someone looks at the code.

python/lsst/utils/tests.py
```diff
@@ -121,13 +122,21 @@ def suiteClassWrapper(tests):
...
         """Reset the leak counter when the tests have been completed"""
         init()
 
     def testFileDescriptorLeaks(self) -> None:
-        """Check if any file descriptors are open since init() called."""
+        """Check if any file descriptors are open since init() called.
+
+        Excludes files with certain known path components and any files
+        that match regexp patterns in ``self.exclude_list``, which can
+        be set in the ``setUp()`` method.
```
timj, line 138: This paragraph is out of date.

timj (approved):
Thanks for doing this.

PR #156, opened 2023-04-21
DM-38812: Prefer using importlib.metadata rather than __version__

timj:
For example jsonschema has decided to no longer support `__version__` and wants to use importlib.metadata.

The downside of this is that importlib.metadata is much slower than asking for the `__version__` of a module that's already imported. Hundreds of milliseconds vs microseconds to get the package versions and importlib.metadata does not seem to cache the result.

## Checklist

- [ ] ran Jenkins


ktlim (approved):
The code itself looks fine, but I do worry a bit about performance (see ticket).

PR #159, opened 2023-05-30
DM-39402: Refactor python package determination to use a hierarchy

timj:
Using `importlib.metadata.version` can be slow so using it for "a.b.c" and "a.b.c.d1" and "a.b.c.d2" when we know the version of "a.b.c" is wasteful. Use the hierarchy to minimize the number of version requests. Also skip standard library modules.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/packages.py
+++ b/python/lsst/utils/packages.py
@@ -124,37 +124,60 @@ def getPythonPackages() -> Dict[str, str]:
 
     # Not iterating with sys.modules.iteritems() because it's not atomic and
     # subject to race conditions
-    module_names = []
-    for name in list(sys.modules.keys()):
-        if name.startswith("_"):
-            # No reason to think we need to report the version number of
-            # private packages.
-            continue
-        if name in _STDLIB:
-            # Assign all standard library packages the python version.
-            packages[name] = sys.version
-        else:
-            module_names.append(name)
+    module_names = list(sys.modules.keys())
 
     # Use knowledge of package hierarchy to find the versions rather than
     # using each name independently. Group all the module names into the
     # hierarchy, splitting on dot, and skipping any component that starts
     # with an underscore.
-    hierarchy: dict[str, dict] = {}
-    for name in module_names:
-        components = name.split(".")
 
-        current = hierarchy
-        for comp in components:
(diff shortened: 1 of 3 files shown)


python/lsst/utils/packages.py
```diff
@@ -114,55 +121,131 @@ def getPythonPackages() -> Dict[str, str]:
...
-
-        # Remove "foo.bar.version" in favor of "foo.bar"
-        # This prevents duplication when the __init__.py includes
-        # "from .version import *"
-        modified = False
-        for ending in (".version", "._version"):
-            if name.endswith(ending):
-                name = name[: -len(ending)]
-                modified = True
+    module_names = []
```
ktlim, line 127: Isn't this slower than the `list()` that was there before to try to limit race conditions, according to the comment?

python/lsst/utils/packages.py
```diff
@@ -114,55 +121,131 @@ def getPythonPackages() -> Dict[str, str]:
...
+            # Assign all standard library packages the python version.
+            packages[name] = sys.version
+        else:
+            module_names.append(name)
+
+    # Use knowledge of package hierarchy to find the versions rather than
+    # using each name independently. Group all the module names into the
+    # hierarchy, splitting on dot, and skipping any component that starts
+    # with an underscore.
+    hierarchy: dict[str, dict] = {}
```
ktlim, line 143: I think you can get the same effect in a way that's perhaps slightly simpler by just sorting the list of packages.  Then as soon as you hit a `_` or known version, skip until you get a different prefix.

timj: Good point. I was possibly over-thinking it.

python/lsst/utils/packages.py
timj: I have to iterate through to assign the standard library versions to the python version.

python/lsst/utils/packages.py
timj: I've rewritten it. It's much simpler now and takes about the same amount of time to execute so it seems like a win.

ktlim (approved):
Looks great.

PR #160, opened 2023-06-05
DM-39410:  Add DeprecationWarning to ellipsis module

andy-slac:
The warning is produced for Python versions that define `types.EllipsisType`
(3.10 or newer). Older Python versions may still need this mypy workaround
and will not see the warning.

## Checklist

- [X] ran Jenkins


timj:
In theory this package still works on python 3.8 so maybe the fallback code should exist even if we no longer use it in the python 3.10-only middleware.

andy-slac:
> In theory this package still works on python 3.8 so maybe the fallback code should exist even if we no longer use it in the python 3.10-only middleware.

Its only use was in `daf_butler`. But I can keep it here, if you prefer.

andy-slac:
OK, restored the module but added a DeprecationWarning.

doc/changes/DM-39410.other.md
```diff
@@ -0,0 +1 @@
+A mypy workaround in `ellipsis` module is not needed for Python 3.10 or newer, importing `lsst.utils.ellipsis` in these Python versions will produce a DeprecationWarning.
```
timj, line 1:
```suggestion
A mypy workaround in `ellipsis` module is not needed for Python 3.10 or newer, importing `lsst.utils.ellipsis` in these Python versions will produce a `DeprecationWarning`.
```

I think we might want this still to be "removal" (since this is an API deprecation) rather than "other" (in fact "other" was removed long ago as an option but I see it didn't get removed from the `doc/changes/README.rst` so that needs fixing.

timj (approved):

PR #162, opened 2023-06-25
DM-39785: Package modernization and ruff

timj:
## Checklist

- [x] ran Jenkins


.github/workflows/lint.yaml
```diff
@@ -9,3 +9,8 @@ on:
 jobs:
   call-workflow:
     uses: lsst/rubin_workflows/.github/workflows/lint.yaml@main
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: chartboost/ruff-action@v1
```
andy-slac, line 16: Should we be adding a reusable workflow for ruff to `rubin_workflows`?

tests/import_test/two/three/success.py
```diff
@@ -1,16 +1,21 @@
...
     return True
 
 
 class Container:
-    def inside():
+    """Class for testing stacklevel."""
+
+    def inside() -> int:
+        """Return 1."""
         return "1"
```
andy-slac, line 14: This returns `str`, not `int`?

tests/test_executables.py
```diff
@@ -16,6 +16,8 @@
 
 
 class ExplicitBinaryTester(lsst.utils.tests.ExecutablesTestCase):
+    """Test that an named executable can be run."""
```
andy-slac, line 19: `a named`?

tests/test_getTempFilePath.py
```diff
@@ -121,9 +126,12 @@ def testClash(self):
 
 
 class TestNameClash2(unittest.TestCase):
+    """Test involving a potential class of test method name."""
```
andy-slac, line 129:
```suggestion
    """Test involving a potential clash of test method name."""
```

tests/test_getTempFilePath.py
```diff
@@ -132,6 +140,8 @@ def testClash(self):
 
 
 class TestNameClash3(unittest.TestCase):
+    """Test involving a potential class of test method name."""
```
andy-slac, line 143:
```suggestion
    """Test involving a potential clash of test method name."""
```

andy-slac (approved):
Looks good, few minor comments.

.github/workflows/lint.yaml
timj: Yes, although I don't think we can add it to the reusable linter workflow so we'd have to make a standalone workflow with those four lines in it.  Maybe a shared linter could include ruff but not run it if a use_ruff parameter is false and then we can add a "use_ruff = True" argument to that. I'm happy to fix things up based on the RFC.

PR #164, opened 2023-07-01
DM-39857: Stop using pytest-flake8 for testing

timj:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,8 +35,6 @@ dynamic = ["version"]
 [project.optional-dependencies]
 test = [
     "pytest >= 3.2",
-    "flake8 >= 3.7.5",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -104,8 +102,6 @@ line_length = 110
 write_to = "python/lsst/utils/version.py"
 
 [tool.pytest.ini_options]
```
ktlim, line 104: You're sure the empty section is still needed/desirable?

ktlim (approved):

pyproject.toml
timj: I can confirm that this configuration will be picked up by pytest. If I take it out pytest will look in parent directories for the config.

PR #165, opened 2023-07-06
DM-35901: Add new test case for importing every python file in a package

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/tests.py
+++ b/python/lsst/utils/tests.py
@@ -32,7 +32,7 @@ __all__ = [
 import contextlib
 import functools
 import gc
-import importlib
+import importlib.resources as resources
 import inspect
 import itertools
 import os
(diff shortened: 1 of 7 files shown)


python/lsst/utils/tests.py
```diff
@@ -331,6 +335,67 @@ def create_executable_tests(cls, ref_file: str, executables: Sequence[str] | Non
...
+                continue
+            if file.startswith("__"):
+                continue
+            root, _ = os.path.splitext(file)
+            module_name = f"{root_pkg}.{root}"
+            with self.subTest(module=module_name):
+                try:
+                    doImport(module_name)
+                except ImportError as e:
+                    raise AssertionError(f"Error importing module {module_name}: {e}") from e
```
erykoff, line 396: Is it possible to test the failing case?  Maybe by writing out a junk python file to a temp directory?

timj: I'm not entirely sure how I would catch the failure since I'm not writing the test that is failing, the infrastructure is writing it and attaching it directly to this class.

erykoff (approved):

PR #169, opened 2023-08-17
DM-40395: Ensure that absolute path is used in getTempFilePath fallback

timj:
Using outdir of "" leads to a relative path which can cause confusion in some scenarios.

## Checklist

- [ ] ran Jenkins


andy-slac (approved):
Looks good

PR #171, opened 2023-09-14
DM-22287: Make inheritDoc more robust and add tests

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/inheritDoc.py
+++ b/python/lsst/utils/inheritDoc.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 __all__ = ("inheritDoc",)
 
+import inspect
 from collections.abc import Callable
 
 
@@ -34,6 +35,18 @@ def inheritDoc(klass: type) -> Callable:
     -------
     decorator : callable
         Intermediate decorator used in the documentation process.
+
+    Notes
+    -----
+    This method naively appends the doc string from the decorated method to the
+    doc string of the equivalent method from the given class. No attempt
+    is made to ensure that duplicated sections are merged together or
+    overwritten.
+
+    This decorator is not necessary to ensure that a parent doc string appears
+    in a subclass. Tools like ``pydoc`` and Sphinx handle that automatically.
+    This can, though, be used to ensure that a specific docstring from a
+    parent class appears if there is ambiguity from multiple inheritance.
     """
 
     def tmpDecorator(method: type) -> Callable:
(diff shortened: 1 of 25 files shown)


python/lsst/utils/inheritDoc.py
```diff
@@ -34,15 +35,41 @@ class and extend it with any additional documentation defined.
     -------
     decorator : callable
         Intermediate decorator used in the documentation process.
+
+    Notes
+    -----
+    This method naively appends the doc string from this method to the
```
andy-slac, line 41:
```suggestion
    This method naively appends the doc string from the decorated method to the
```
?

python/lsst/utils/inheritDoc.py
```diff
@@ -34,15 +35,41 @@ class and extend it with any additional documentation defined.
...
+
+        if parentText and appendText:
+            # cleandoc() strips leading and trailing space so it is safe
+            # to add new lines.
+            method.__doc__ = parentText + "\n\n" + appendText
+        elif parentText:
+            method.__doc__ = parentText
+        else:
+            # Do not update the doc string if there was no parent doc string
+            # and no method doc string.
```
andy-slac, line 71:
```suggestion
            # Do not update the doc string if there was no parent doc string.
```
?

tests/test_inheritDoc.py
```diff
@@ -0,0 +1,122 @@
...
+        self.assertEqual(
+            inspect.getdoc(InheritDoc.method2),
+            """Return Method 2.
+
+Note on method 2.""",
+        )
+        self.assertEqual(inspect.getdoc(InheritDoc.method3), "Note on method 3.")
+        self.assertEqual(inspect.getdoc(InheritDoc.method4), "Return Method 4.")
+        self.assertIsNone(inspect.getdoc(InheritDoc.method5))
+
```
andy-slac, line 119: Maybe add a test case for methods with oddly indented docstrings?

andy-slac (approved):
Looks good, a couple of minor comments.

python/lsst/utils/inheritDoc.py
timj: I was trying to say that if you have no docstring in your method and there is no docstring in the parent, this decorator won't change the contents of `__doc__`. Before `inheritDoc` would change it to an empty string but now I leave it as `None`... Maybe that's assumed?

tests/test_inheritDoc.py
timj: What sort of odd?

This:

```python
def method(self):
    """Doc string

        Indented oddly.
    Indented normally.
    ""
```
? Which should leave the "Indented oddly" line still indented.

andy-slac: Maybe
```
def method(self):
    """
    Doc string

    Indented normally."""
```
?

python/lsst/utils/inheritDoc.py
andy-slac: The `else` branch is for the case when parent `__doc__` is empty or None, and it does not check whether method doc is empty. I think it's fine to not update it, I was just checking that logic matches the comment.

timj: You are right. Maybe that would be clearer if I rewrote as:

```python
if parentText:
    if appendText:
        ...
    else:
        ...
else:
    pass
```

tests/test_inheritDoc.py
timj: Okay. That is actually one of the cases in `pipe_tasks` and it turns out that `cleandoc` deals with it and removes all the leading white space. Will add.

PR #173, opened 2023-09-18
DM-40817: Allow skipping modules in ImportTestCase

arunkannawadi:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/tests.py
+++ b/python/lsst/utils/tests.py
@@ -43,7 +43,12 @@ import tempfile
 import unittest
 import warnings
 from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
-from importlib import resources
+
+if sys.version_info < (3, 10, 0):
+    import importlib_resources as resources
+else:
+    from importlib import resources
+
 from typing import Any, ClassVar
 
 import numpy
--- a/python/lsst/utils/tests.py
+++ b/python/lsst/utils/tests.py
@@ -42,7 +42,7 @@ import sys
 import tempfile
 import unittest
 import warnings
-from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
+from collections.abc import Callable, Container, Iterable, Iterator, Mapping, Sequence
 
 if sys.version_info < (3, 10, 0):
     import importlib_resources as resources
@@ -351,6 +351,16 @@ class ImportTestCase(unittest.TestCase):
     PACKAGES: ClassVar[Iterable[str]] = ()
     """Packages to be imported."""
 
+    SKIP_FILES: ClassVar[Mapping[str, Container[str]]] = {}
+    """Files to be skipped importing; specified as key-value pairs.
+
+    The key is the package name and the value is a set of files names in that
+    package to skip.
+
+    Note: Files with names not ending in .py or beginning with leading double
+    underscores are always skipped.
+    """
+
     _n_registered = 0
     """Number of packages registered for testing by this class."""
 
@@ -388,9 +398,13 @@ class ImportTestCase(unittest.TestCase):
     def assertImport(self, root_pkg):
         for file in resources.files(root_pkg).iterdir():
             file = file.name
+            # When support for python 3.9 is dropped, this could be updated to
+            # use match case construct.
             if not file.endswith(".py"):
                 continue
             if file.startswith("__"):
+                continue
+            if file in self.SKIP_FILES.get(root_pkg, ()):
                 continue
             root, _ = os.path.splitext(file)
             module_name = f"{root_pkg}.{root}"
--- a/tests/test_import.py
+++ b/tests/test_import.py
@@ -38,5 +38,16 @@ class TestImports(ImportTestCase):
         self.assertEqual(self._n_registered, 1)
 
 
+class TestSkipImports(ImportTestCase):
+    """Test that we can run the import tests with some modules skipped."""
+
+    PACKAGES = ("import_test.two.three",)
+    SKIP_FILES = {"import_test.two.three": {"runtime.py", "fail.py"}}
+
+    def test_import(self):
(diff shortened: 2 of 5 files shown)


arunkannawadi:
Do we actually support python 3.9? There's a strange error that doesn't occur with python 3.10 or 3.11 and I am not sure if I should dig deep into this.

timj:
The error on 3.9 seems to be an interaction between `importlib.resources` and how it scans that package that lives in the `tests` directory.  I think we have three options:

* Change tests.py to use `importlib_resources` on python 3.9 (which is the 3.11 version of the code) and add dependency.
* Give up on 3.9 completely
* Skip the test case on python 3.9.

None of these options are more than a few lines to fix. My philosophy for utils is that, given it's a very low level package, I try to support as old a python as I can until I get annoyed by doing so.

timj:
I think my preference is to see if importlib_resources works on 3.9.

```python
import sys
if sys.version_info < (3, 10, 0):
    import importlib_resources as resources
else:
    from importlib import resources
```
or something and see if that fixes it. (will need importlib_resources added as dependency to pyproject.toml and requirements.txt and I don't think there is a way to only make that happen on 3.9 distros)

tests/test_import.py
```diff
@@ -38,5 +38,16 @@ def test_counter(self):
...
 
+class TestSkipImports(ImportTestCase):
+    """Test that we can run the import tests with some modules skipped."""
+
+    PACKAGES = ("import_test.two.three",)
+    SKIP_MODULES = {"import_test.two.three": {"runtime.py", "fail.py"}}
+
+    def test_import(self):
+        """Test that we can import utils."""
+        import import_test  # noqa: F401
```
timj, line 49: This import might be a no-op depending on what other tests have already run.

python/lsst/utils/tests.py
```diff
@@ -346,6 +346,12 @@ class ImportTestCase(unittest.TestCase):
     PACKAGES: ClassVar[Iterable[str]] = ()
     """Packages to be imported."""
 
+    SKIP_MODULES: ClassVar[Mapping[str, Iterable[str]]] = {}
+    """Modules to be skipped importing as key-value pairs.
+
+    The key is the package name and the value is a set of module names to skip.
```
timj, line 352: Is "module name" the right word here? Aren't these file names? ie you are requiring the `.py` extension.

python/lsst/utils/tests.py
```diff
@@ -387,6 +393,8 @@ def assertImport(self, root_pkg):
                 continue
             if file.startswith("__"):
                 continue
+            if self.SKIP_MODULES and file in self.SKIP_MODULES.get(root_pkg, ()):
```
timj, line 400: Are you worried that someone might have written the test with a value of `SKIP_MODULES = None` here? I think you can likely ignore that possibility since the developer writing the code will find out immediately when it crashes and we don't put a similar protection around `cls.PACKAGES`.

python/lsst/utils/tests.py
```diff
@@ -43,7 +43,11 @@
 import unittest
 import warnings
 from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
-from importlib import resources
+import sys
```
timj, line 46: Remove.

python/lsst/utils/tests.py
```diff
@@ -43,7 +43,11 @@
...
+if sys.version_info < (3, 10, 0):
+    import importlib_resources as resources
```
timj, line 48: I think that `importlib_resources` has to be added as a dependency to pyproject.toml and requirements.txt

timj (approved):
Some minor comments.

python/lsst/utils/tests.py
arunkannawadi: `from importlib import resources` seems to work in `python3.9` anyway. Does that have a different behavior despite the same name?

python/lsst/utils/tests.py
arunkannawadi: This was all when I thought something being `None` here would be triggering the failure in py3.9. I will just switch to

```python
if file in self.SKIP_MODULES.get(root_pkg, ()):
```

or whatever name we settle on.

python/lsst/utils/tests.py
timj: It's not the same name. It's different. `importlib_resources` is a PyPI package that allows older pythons to use the newer implementation of `importlib.resources`. That's why I was saying that you need to add `importlib_resources` as a dependency in pyproject.toml and requirements.txt.

timj: `importlib.resources` was working fine on python 3.9 until you caused it trouble by asking it to scan that package inside the tests directory. The reason I'm asking you to try `importlib_resources` on python 3.9 is that they might have fixed a bug in that area which explains why it worked fine on 3.10.

python/lsst/utils/tests.py
arunkannawadi: I could change it to `SKIP_FILES`. While on that, I should also change the type of the value to be `Container[str]`, since I only need `in` to work.

python/lsst/utils/tests.py
arunkannawadi: Okay, this patch works! Thanks Tim, will clean up the changes, address the other minor comments and merge.

python/lsst/utils/tests.py
```diff
@@ -346,6 +351,12 @@ class ImportTestCase(unittest.TestCase):
     PACKAGES: ClassVar[Iterable[str]] = ()
     """Packages to be imported."""
 
+    SKIP_FILES: ClassVar[Mapping[str, Container[str]]] = {}
```
timj, line 354: Don't forget to change the doc string.

PR #178, opened 2024-01-08
DM-42391: Make use of packages_distributions() in utils.packages

erykoff:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/packages.py
+++ b/python/lsst/utils/packages.py
@@ -50,7 +50,7 @@ BUILDTIME = {"boost", "eigen", "tmv"}
 # Python modules to attempt to load so we can try to get the version
 # We do this because the version only appears to be available from python,
 # but we use the library
-PYTHON = {"galsim"}
+PYTHON = {}
 
 SPECIAL_NAMESPACES = {"lsst"}
 
--- a/python/lsst/utils/packages.py
+++ b/python/lsst/utils/packages.py
@@ -28,6 +28,7 @@ import sys
 import types
 from collections.abc import Mapping
 from functools import lru_cache
+from importlib.metadata import packages_distributions
 from typing import Any, ClassVar
 
 import yaml
@@ -50,6 +51,8 @@ BUILDTIME = {"boost", "eigen", "tmv"}
 # We do this because the version only appears to be available from python,
 # but we use the library
 PYTHON = {"galsim"}
+
+SPECIAL_NAMESPACES = {"lsst"}
 
 # Packages that don't seem to have a mechanism for reporting the runtime
 # version.  We need to guess the version from the environment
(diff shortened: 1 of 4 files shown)


erykoff:
Also on this PR I removed support for python 3.9 (since `packages_distributions()` requires 3.10), and pinned the documenteer version to 0.8.2.

python/lsst/utils/packages.py
```diff
@@ -49,7 +50,9 @@
 # Python modules to attempt to load so we can try to get the version
 # We do this because the version only appears to be available from python,
 # but we use the library
-PYTHON = {"galsim"}
+PYTHON: set[str] = set()
```
timj, line 53: In some sense we should delete all this stuff since it's a bit unreliable and for us conda reports all these versions directly. Not for this ticket of course.

python/lsst/utils/packages.py
```diff
@@ -165,8 +170,19 @@ def getPythonPackages() -> dict[str, str]:
             # for the line after this.
             continue
 
-        # Look for a version.
-        ver = _get_python_package_version(name, packages)
+        # Find the namespace which we need to use package_dist.
+        namespace = name.split(".")[0]
+
+        # package_dist is a mapping from import namespace to distribution
+        # package names. This may be a one-to-many mapping due to namespace
```
timj, line 177: Such as:

```
sphinxcontrib: ['sphinxcontrib-bibtex', 'sphinxcontrib-autoprogram', 'sphinxcontrib-htmlhelp', 'sphinxcontrib-qthelp', 'sphinxcontrib-serializinghtml', 'sphinxcontrib-devhelp', 'sphinxcontrib-jsmath', 'sphinxcontrib-applehelp', 'sphinxcontrib-doxylink', 'sphinxcontrib-jquery']
lsst: ['ts-xml']
```
?

python/lsst/utils/packages.py
```diff
@@ -210,10 +232,35 @@ def _get_python_package_version(name: str, packages: dict[str, str]) -> str | No
...
+        # Split the name into parts.
+        name_parts = re.split("[._-]", name)
+
+        found = False
+        for dist_name in dist_names:
+            dist_name_parts = re.split("[._-]", dist_name)
+            # Check if the components start with the namespace; this is
+            # needed because (at least) lsst.ts packages do not use
+            # ``lsst`` in the package name.
+            if dist_name_parts[0] != namespace:
```
timj, line 247: Do you want to install ts-xml in the action so you can get code coverage? It would be nice if we always had the `lsst` prefix in the public distributions of things like this...

python/lsst/utils/packages.py
```diff
@@ -210,10 +232,35 @@ def _get_python_package_version(name: str, packages: dict[str, str]) -> str | No
...
+                dist_name_parts.insert(0, namespace)
+
+            if dist_name_parts == name_parts:
+                found = True
+                break
+
+        if not found:
+            # Fall back to name.
+            dist_name = name
```
timj, line 256: Is this what happens currently in the EUPS case? Maybe mention that in the comment?

timj (approved):
Thanks for doing this.

For the attrs/attr case do we still end up requesting the version from `attr` twice? Or is the second time skipped because of cache access?

python/lsst/utils/packages.py
erykoff: Exactly.  This is somewhat described in https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#packaging-namespace-packages but it's not strictly required but it is the convention for every set that I've seen.

erykoff:
Re: `attr`/`attrs`.  The `attrs` package supplies both `attr` and `attrs` and so each of these gets listed in the imported modules if they've both been imported.  And each gets its own version lookup (but just once each).  I don't know if the package metadata lookup is cached or not in this situation.

PR #183, opened 2024-06-07
DM-44725: Add make_figure() utility to create non-interactive matplotlib Figures

erykoff:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/doc/changes/DM-44725.md
+++ b/doc/changes/DM-44725.md
@@ -1,0 +1,1 @@
+Add `lsst.utils.plotting.make_figure()` utility to create a non-interactive `matplotlib` `Figure` without `pyplot`.
--- a/python/lsst/utils/plotting/__init__.py
+++ b/python/lsst/utils/plotting/__init__.py
@@ -11,3 +11,4 @@
 """General LSST Plotting Utilities."""
 
 from .figures import *
+from .limits import *
--- a/python/lsst/utils/plotting/__init__.py
+++ b/python/lsst/utils/plotting/__init__.py
@@ -1,0 +1,13 @@
+# This file is part of utils.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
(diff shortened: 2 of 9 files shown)


doc/changes/DM-44725.md
```diff
@@ -0,0 +1 @@
+Add lsst.utils.plotting.make_figure() utility to create non-interactive matplotlib Figure without pyplot.
```
timj, line 1:
```suggestion
Add `lsst.utils.plotting.make_figure()` utility to create a non-interactive `matplotlib` `Figure` without `pyplot`.
```

python/lsst/utils/plotting/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# Use of this source code is governed by a 3-clause BSD-style
+# license that can be found in the LICENSE file.
+"""General LSST Plotting Utilities."""
+
+from .figures import *
```
timj, line 13: We should probably add `from .limits import *` as well and add a `__all__` to `limits.py`.

timj (approved):
Looks fine. It did make me notice that `plotImageDiff` in `lsst.utils.tests` which should clearly now move to the `plotting` package and try to not use pyplot internally. Nothing in our code is using it anyhow.

timj (approved):

gpdf:
I only noticed this after the fact.  I don't fully understand the "optional dependencies" issue here, but it worries me that this low-level package, which is (I think) one of the few dependencies of the whole Butler/PipelineTask stack, seems to now have a non-test-limited `matplotlib` dependency.

erykoff:
I don't understand what you mean by "non-test-limited matplotlib dependency".  It's an optional dependency.  And note that there was always matplotlib code here, it wasn't declared (or tested).

timj:
We have set it up so that `pip install utils` will not install matplotlib but `pip install utils[plotting]` will install matplotlib. Ideally I don't want matplotlib code in there at all but we didn't have time to RFC a new plotutils package.

PR #184, opened 2024-07-09
DM-45007: Add simple memory profiling tools

kfindeisen:
This PR adds a couple of simple tools that were developed for investigating [DM-45007](https://rubinobs.atlassian.net/browse/DM-45007).

## Checklist

- [x] ran Jenkins

[DM-45007]: https://rubinobs.atlassian.net/browse/DM-45007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/changes/DM-45007.feature.md
+++ b/doc/changes/DM-45007.feature.md
@@ -1,0 +1,1 @@
+Added `lsst.utils.introspection.take_object_census` function for characterizing memory leaks.
--- a/doc/changes/DM-45007.feature.md
+++ b/doc/changes/DM-45007.feature.md
@@ -1,1 +1,1 @@
-Added `lsst.utils.introspection.take_object_census` function for characterizing memory leaks.
+Added `lsst.utils.introspection.take_object_census` and `lsst.utils.introspection.trace_object_references` functions for characterizing memory leaks.
--- a/python/lsst/utils/introspection.py
+++ b/python/lsst/utils/introspection.py
@@ -20,12 +20,14 @@ __all__ = [
     "get_caller_name",
     "find_outside_stacklevel",
     "take_object_census",
+    "trace_object_references",
 ]
 
 import builtins
 import collections
 import gc
 import inspect
+import itertools
 import sys
 import types
 import warnings
@@ -237,7 +239,7 @@ def find_outside_stacklevel(
 
     Examples
     --------
-    .. code-block :: python
+    .. code-block:: python
 
         warnings.warn(
             "A warning message",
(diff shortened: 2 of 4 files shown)


python/lsst/utils/introspection.py
```diff
@@ -301,3 +306,127 @@ def find_outside_stacklevel(
...
+    -------
+    refs : `list`
+        The objects that refer to the elements of ``objects``, not counting
+        ``objects`` itself.
+    """
+    refs = gc.get_referrers(*objects)
+    refs.remove(objects)
+    iterators = [x for x in refs if type(x).__name__.endswith("_iterator")]
+    for i in iterators:
+        refs.remove(i)
```
timj, line 431: Do you understand why `iterators` is always empty in your test code?

python/lsst/utils/introspection.py
```diff
@@ -301,3 +306,127 @@ def find_outside_stacklevel(
...
+        that is known to be leaking.
+    count : `int`
+        The number of example objects to trace, if that many exist.
+    max_level : `int`
+        The number of levels of references to trace. ``max_level=1`` means
+        finding only objects that directly refer to the examples.
+
+    Returns
+    -------
+    traces : `list` [collection]
```
timj, line 355: Could this be tuple rather than list? Or is the same list extended and reused during the recursion?

`[collection]` is confusing to me. Do you mean `collections.abc.Sequence`?

python/lsst/utils/introspection.py
```diff
@@ -301,3 +306,127 @@ def find_outside_stacklevel(
...
+        A sequence whose first element (index 0) is the set of example objects
```
timj, line 356: This is complicated enough to describe that it might be nice to have an Examples section with explicit outputs.

python/lsst/utils/introspection.py
```diff
@@ -301,3 +306,127 @@ def find_outside_stacklevel(
...
+        ``max_level + 1`` elements.
+    trace_complete : `bool`
+        `True` if the trace for all objects terminated in at most
+        ``max_level`` references, and `False` if more references exist.
+    """
+
+    def class_filter(o: Any) -> bool:
+        return isinstance(o, target_class)
+
+    # set would be more appropriate, but objects may not be hashable
```
timj, line 368:
```suggestion
    # set() would be more appropriate, but objects may not be hashable.
```

doc/changes/DM-45007.feature.md
```diff
@@ -0,0 +1 @@
+Add `utils.introspection.take_object_census` and `~utils.introspection.trace_object_references` functions for characterizing memory leaks.
```
timj, line 1:
```suggestion
Added `lsst.utils.introspection.take_object_census` and `lsst.utils.introspection.trace_object_references` functions for characterizing memory leaks.
```

timj (approved):
Looks okay. Some minor comments but there are no tests for one of the new functions.

python/lsst/utils/introspection.py
```diff
@@ -301,3 +306,127 @@ def find_outside_stacklevel(
...
+
+    Notes
+    -----
+    This function counts _all_ Python objects in memory. To count only
+    reachable objects, run `gc.collect` first.
+    """
+    counts: collections.Counter[type] = collections.Counter()
+    for obj in gc.get_objects():
+        counts[type(obj)] += 1
+    return counts
```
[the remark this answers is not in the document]
kfindeisen: As noted in the commit message, I have no idea how to test it. Do you have any suggestions?

python/lsst/utils/introspection.py
kfindeisen: I do not consider `tuple` to be an appropriate API type here -- this is a sequence of alike things (the relationship between the ith and i-1st entries is fixed), not a generalized argument list.

As for the type, I forgot to change the doc after MyPi pointed out that a `list[list]` is not a `list[Collection]`. It should be `list[list]`.

python/lsst/utils/introspection.py
kfindeisen: Unfortunately, no. When I was first writing/editing this (in the PP code) there was always a `list_iterator` that pointed to the objects and introduced a circular reference to the working list. I assumed it was implicitly called by the `*` operator, and don't know why it doesn't show up here.

python/lsst/utils/introspection.py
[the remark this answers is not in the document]
timj: Can you call it in a test to at least ensure it doesn't crash and returns something of the right type?

PR #186, opened 2024-07-22
DM-43599: Clean up docs for PeriodicLogger

kfindeisen:
This PR makes some corrections and clarifications to the docs for `PeriodicLogger`, to make it easier to use.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/utils/logging.py
+++ b/python/lsst/utils/logging.py
@@ -382,6 +382,9 @@ class PeriodicLogger:
     be useful to issue a log message periodically to show that the
     algorithm is progressing.
 
+    The first time threshold is counted from object construction, so in general
+    the first call to `log` does not log.
+
     Parameters
     ----------
     logger : `logging.Logger` or `LsstLogAdapter`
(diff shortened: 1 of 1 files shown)


python/lsst/utils/logging.py
```diff
@@ -413,12 +416,16 @@ def __init__(self, logger: LsstLoggers, interval: float | None = None, level: in
...
+        creation of this object.
+
         Parameters
         ----------
         msg : `str`
             Message to issue if the time has been exceeded.
         *args : Any
-            Parameters to be passed to the log system.
+            Arguments to be merged into the message string, as described under
+            `logging.Logger.debug`.
```
timj, line 428: Shouldn't this be `logging.Logger.log` rather than debug?

kfindeisen: No, [`Logger.debug`](https://docs.python.org/3/library/logging.html#logging.Logger.debug) is what actually explains the arguments. [`Logger.log`](https://docs.python.org/3/library/logging.html#logging.Logger.log) does not contain that information, and you have to follow two more links to get to it. And the point, after all, is for the reader to understand what they can or can't pass in here.

bsmartradio (approved):
Looks good to me.

PR #198, opened 2025-01-15
DM-48430: Add ability to get all installed packages from system

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-48430.feature.rst
+++ b/doc/changes/DM-48430.feature.rst
@@ -1,0 +1,3 @@
+* Add new ``lsst.utils.packages.getAllPythonDistributions`` function for retrieving the versions of all installed Python distributions regardless of what has been imported.
+* Modified ``Packages.fromSystem`` to add a flag, ``include_all``, to control whether all installed Python distributions and EUPS packages should be reported and not solely the (default) imported distributions and locally-setup EUPS packages. All installed Conda packages are always reported.
+* Modified ``lsst.utils.packages.getPythonPackages`` to no longer report packages that are part of the Python standard library. It is sufficient to record the Python version (via the ``python`` key).
--- a/python/lsst/utils/packages.py
+++ b/python/lsst/utils/packages.py
@@ -344,7 +344,7 @@ def getEnvironmentPackages(include_all: bool = False) -> dict[str, str]:
         if not prod.version.startswith(Product.LocalVersionPrefix):
             if include_all:
                 tags = {t for t in prod.tags if t != "current"}
-                tag_msg = " (" + " ".join(tags) + ")" if tags else ""
+                tag_msg = " (" + " ".join(sorted(tags)) + ")" if tags else ""
                 packages[prod.name] = prod.version + tag_msg
             continue
         ver = prod.version
--- a/python/lsst/utils/packages.py
+++ b/python/lsst/utils/packages.py
@@ -140,6 +140,9 @@ def getPythonPackages() -> dict[str, str]:
     module.  Note, therefore, that we can only report on modules that have
     *already* been imported.
 
+    Python standard library packages are not included in the report. A
+    ``python`` key is inserted that records the Python version.
+
     We don't include any module for which we cannot determine a version.
 
     Whilst distribution names are used to determine package versions, the
@@ -187,13 +190,6 @@ def getPythonPackages() -> dict[str, str]:
             # packages such as _abc and __future__.
             continue
 
-        if name in _STDLIB:
-            # Assign all standard library packages the python version
-            # since they almost all lack explicit versions.
-            packages[name] = sys.version
-            previous = name
-            continue
-
         if name.startswith(previous + ".") and previous in packages:
             # Already have this version. Use the same previous name
             # for the line after this.
@@ -201,6 +197,13 @@ def getPythonPackages() -> dict[str, str]:
 
         # Find the namespace which we need to use package_dist.
         namespace = name.split(".")[0]
+
+        if namespace in _STDLIB:
+            # If this is an import from the standard library, skip it.
+            # Standard library names only refer to top-level namespace
+            # so "importlib" appears but "importlib.metadata" does not.
+            previous = name
+            continue
 
         # package_dist is a mapping from import namespace to distribution
         # package names. This may be a one-to-many mapping due to namespace
--- a/python/lsst/utils/packages.py
+++ b/python/lsst/utils/packages.py
@@ -196,15 +196,24 @@ def getPythonPackages() -> dict[str, str]:
         n_versions,
     )
 
+    return _mangle_lsst_package_names(packages)
+
+
+def _mangle_lsst_package_names(packages: dict[str, str]) -> dict[str, str]:
     for name in list(packages.keys()):
         # Use LSST package names instead of python module names
         # This matches the names we get from the environment (i.e., EUPS)
         # so we can clobber these build-time versions if the environment
         # reveals that we're not using the packages as-built.
         if name.startswith("lsst."):
-            new_name = name.replace("lsst.", "").replace(".", "_")
-            packages[new_name] = packages[name]
-            del packages[name]
+            sep = "."
+        elif name.startswith("lsst-"):
+            sep = "-"
+        else:
+            continue
+        new_name = name.replace(f"lsst{sep}", "").replace(sep, "_")
+        packages[new_name] = packages[name]
+        del packages[name]
 
     return packages
 
--- a/python/lsst/utils/packages.py
+++ b/python/lsst/utils/packages.py
@@ -27,7 +27,7 @@ import subprocess
 import sys
 import types
 from collections.abc import Mapping
-from functools import lru_cache
+from functools import cache, lru_cache
 from importlib.metadata import packages_distributions
 from typing import Any, ClassVar
 
@@ -37,6 +37,7 @@ log = logging.getLogger(__name__)
 
 __all__ = [
     "getVersionFromPythonModule",
+    "getAllPythonDistributions",
     "getPythonPackages",
     "getEnvironmentPackages",
     "getCondaPackages",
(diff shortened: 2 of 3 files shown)


timj:
@kfindeisen can you give this a try please? One thing that might be an issue is that the "include_all" option historically includes all the EUPS build numbers. I'm not sure what my motivation for adding that was in #117 but it's an easy fix to remove the tags if we decide they get in the way.

timj:
Reading the Jira ticket I think @mfisherlevine  wanted the weekly tag to show up in the version history. The problem is that it also reports build tags:
```
testdata_cfht gfc90949391 (b7235 b7237 b7238 b7239 b7240 b7241 b7242 b7243 b7244 b7245 b7246 b7247 b7248 b7249 b7250 b7251 b7252 b7253 b7254 b7255 b7256 b7257 b7258 b7259 b7260 b7261 b7262 b7263 b7264 b7265 b7266 b7267 b7268 b7269 b7270 b7272 b7273 b7274 b7275 b7277 b7278 b7279 b7282 b7283 b7284 b7285 b7286 b7287
```

kfindeisen:
Well, it is faster (0.5 s vs. 2.1 s on S3DF). I'm not sure how to test its robustness, though -- I can't run PP without a container, and the `lru_cache` would suppress any changes anyway. I'm a bit worried about what would happen if that decorator is ever removed; if that's supposed to be a core part of the method's behavior, can you document it as such?

mfisherlevine:
As ugly as the build tags are there, I don't think I have anything that _relies_ on the weekly being there alone, and if I need it, I can always try adding a regex of some sort, so I think that's fine (at least for me).

python/lsst/utils/packages.py
```diff
@@ -101,6 +102,30 @@ def getVersionFromPythonModule(module: types.ModuleType) -> str:
     return str(version)
 
 
+@lru_cache(maxsize=1)
```
kfindeisen, line 105: Any reason for using this as opposed to `functools.cache`? I would expect the latter to be more lightweight when there's a bounded number of inputs.

python/lsst/utils/packages.py
```diff
@@ -101,6 +102,30 @@ def getVersionFromPythonModule(module: types.ModuleType) -> str:
...
+def getAllPythonDistributions() -> dict[str, str]:
+    """Get the versions for all Python distributions that are installed.
+
+    Returns
+    -------
+    packages : `dict`
+        Keys (type `str`) are distribution names; values (type `str`) are their
```
kfindeisen, line 112: Per [the dev guide](https://developer.lsst.io/python/numpydoc.html#py-docstring-parameter-types-dict):
```suggestion
    packages : `dict` [`str`, `str`]
        Keys are distribution names; values are their
```

python/lsst/utils/packages.py
```diff
@@ -196,15 +221,24 @@ def getPythonPackages() -> dict[str, str]:
         n_versions,
     )
 
+    return _mangle_lsst_package_names(packages)
+
+
+def _mangle_lsst_package_names(packages: dict[str, str]) -> dict[str, str]:
```
kfindeisen, line 227: In the future, please put this sort of refactoring [on its own commit](https://developer.lsst.io/work/flow.html#git-commit-organization-best-practices); it's less confusing that way.

python/lsst/utils/packages.py
```diff
@@ -292,7 +326,7 @@ def _get_python_package_version(
 _eups: Any | None = None  # Singleton Eups object
 
 
-@lru_cache(maxsize=1)
+@lru_cache(maxsize=2)
```
kfindeisen, line 329: This bugfix should also be its own commit.

python/lsst/utils/packages.py
```diff
@@ -494,22 +531,35 @@ def __setstate__(self, state: dict[str, Any]) -> None:
         self.update(state["_packages"])
 
     @classmethod
-    def fromSystem(cls) -> Packages:
+    def fromSystem(cls, include_all: bool = False) -> Packages:
```
kfindeisen, line 534: And this new feature, too...

tests/test_packages.py
```diff
@@ -173,6 +183,10 @@ def testPackages(self):
             some_yaml = b"list: [1, 2]"
             lsst.utils.packages.Packages.fromBytes(some_yaml, "yaml")
 
+        # Check that "all" packages runs and does not include stdlib.
+        all = lsst.utils.packages.Packages.fromSystem(include_all=True)
+        self.assertNotIn("os", all)
```
kfindeisen, line 188: I don't think this test is correct. The spec says "all installed Python packages will be reported".

kfindeisen (changes requested):
I'm a bit concerned about the new unit test for `fromSystem`, since it requires the exact opposite of what the documentation says. Which is the intended behavior?

Also, please document the new features and bugfixes in `doc/changes`.

tests/test_packages.py
timj: It has to mean "distributions" for include_all -- I will tweak the documentation. `os` is not a distribution, it's a package in the standard library.

kfindeisen: That would mean that `include_all` _removes_ packages from the list, behavior I imagine the user will find very confusing even if it's technically documented...

timj: One option is to remove python standard library from `getPythonPackages` (they have an explicit code path at the moment that recognizes them). An issue is that `getPythonPackages` initially started life as a "give me versions of all packages that were imported" but then became "give me versions of all distributions that had any code imported from them" but left behind the standard library imports even though they weren't distributions. @erykoff  made the change to distributions to work around all the oddities of `yaml` coming from something call `pyyaml` and the version of the package is not really relevant.

`getAllPythonDistributions` is explicitly just worrying about distributions (one of the reasons it is faster than `getPythonPackages`) so standard library is meaningless (we add a special `python` key).

@ktlim do you have an opinion?

kfindeisen: If we're in a hurry, another option is to just drop `fromSystem(include_all)` as a feature -- it's not needed for the actual ticket, and it would reduce confusion between the scope of `getAllPythonDistributions` (non-standard packages, imported or not) and `Packages.fromSystem` (all packages, but only imported).

python/lsst/utils/packages.py
timj: No. I forgot about it.

python/lsst/utils/packages.py
timj: Yes. Copy and paste from another location that used the old style. Will fix.

python/lsst/utils/packages.py
timj: I wanted this to be part of the refactor because it only needed to be its own function when there were two users. I will make it a separate commit.

tests/test_packages.py
timj: @kfindeisen you would be okay calling the conda, python, and eups parts of `fromSystem` individually?

Listing the standard library packages has always been up for debate and it was only in python 3.10 that there was a builtin API for knowing if a package was part of the standard library (which is why previously they were included if they had a separate version but skipped otherwise).

ktlim: Overall, it seems that including stdlib packages that are always present and versioned along with Python seems redundant.  But I am a little concerned that these versions are already baked into datasets so that different results may cause conflicts.  On the other hand, I think any such conflicts would be overwhelmed by the other changes in the release that contains this code, so maybe that isn't really an issue after all.

timj: I don't think we have to worry about older package YAML files in butler repos because by definition this is going to change the version of utils which will trigger all the version mismatch paths.

I think I'm hearing that removing stdlib from getPythonPackages is the thing that people want to happen so I will do that and note it in the release notes.

timj: @kfindeisen I have made the change and triggered a new review request for you.

tests/test_packages.py
```diff
@@ -54,6 +54,7 @@ def testEnvironment(self):
         can do is test that this doesn't fall over.
         """
         lsst.utils.packages.getEnvironmentPackages()
+        lsst.utils.packages.getEnvironmentPackages(include_all=True)
```
kfindeisen, line 73: I don't understand this change, and the commit message is... vague.

doc/changes/DM-48430.feature.rst
```diff
@@ -0,0 +1,2 @@
+* Add new ``lsst.utils.packages.getAllPythonDistributions`` function for retrieving the versions of all installed Python distributions regardless of what has been imported.
+* Modified ``Packages.fromSystem`` to add a flag, ``include_all`` to control whether all installed Python distributions and EUPS packages should be reported and not solely the (default) imported distributions and locally-setup EUPS packages. All installed Conda packages are always reported.
```
kfindeisen, line 2:
```suggestion
* Modified ``Packages.fromSystem`` to add a flag, ``include_all``, to control whether all installed Python distributions and EUPS packages should be reported and not solely the (default) imported distributions and locally-setup EUPS packages. All installed Conda packages are always reported.
```

python/lsst/utils/packages.py
```diff
@@ -558,10 +558,10 @@ def fromSystem(cls, include_all: bool = False) -> Packages:
...
         Note
         ----
-        Standard library imports are included if ``include_all`` is `False`
-        (where `getPythonPackages` is called). Since they are not Python
-        distributions they are not included if ``include_all`` is `True` (where
-        `getAllPythonDistributions` is called).
+        The names of Python distributions can differ from the names of the
+        Python packages installed by those distributions. Since ``include_all``
+        set to `True` uses Python distributions and `False` uses Python
+        packages do not expect that the answers are directly comparable.
```
kfindeisen, line 567: I recommend that this be squashed onto "Add include_all flag...".

(👍 for the rewrite, though.)

python/lsst/utils/packages.py
```diff
@@ -140,6 +140,9 @@ def getPythonPackages() -> dict[str, str]:
     module.  Note, therefore, that we can only report on modules that have
     *already* been imported.
 
+    Python standard library packages are not included in the report. A
+    ``python`` key is inseerted that records the Python version.
```
kfindeisen, line 144:
```suggestion
    ``python`` key is inserted that records the Python version.
```

kfindeisen (approved):
Looks much better, thanks! I won't speculate on why there are now 14 commits instead of a handful.

tests/test_packages.py
timj: I had noticed that the include_all code path wasn't explicitly tested so added the simple test. Granted it will be tested by the fromSystem test later on.

timj:
Thanks. I'll squash a bit now that we got agreement on the approach.

PR #199, opened 2025-01-30
DM-48600: Add helper function for parsing metadata added by timeMethod

isullivan:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/utils/timer.py
+++ b/python/lsst/utils/timer.py
@@ -10,8 +10,7 @@
 # license that can be found in the LICENSE file.
 #
 
-"""Utilities for measuring execution time.
-"""
+"""Utilities for measuring execution time."""
 
 from __future__ import annotations
 
--- a/python/lsst/utils/timer.py
+++ b/python/lsst/utils/timer.py
@@ -14,7 +14,7 @@
 
 from __future__ import annotations
 
-__all__ = ["profile", "logInfo", "timeMethod", "time_this"]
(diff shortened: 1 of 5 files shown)


python/lsst/utils/timer.py
```diff
@@ -480,3 +479,33 @@ def profile(filename: str | None, log: LsstLoggers | None = None) -> Iterator[cP
     profile.dump_stats(filename)
     if log is not None:
         log.info("cProfile stats written to %s", filename)
+
+
+def durationFromTimeMethod(
```
timj, line 484: Please use underscores in names and variables for new functions in this package.

isullivan: @timj Since this is a helper function that works with the results of `timeMethod`, is the name `duration_from_timeMethod` OK, or should I just change it to something like `calculate_duration`?

timj: I can live with `duration_from_timeMethod`. I should have fixed up the name `timeMethod` years ago when I migrated the code to the new location. I've been trying to think of other ways to organize this but we are sort of stymied by the historical baggage of timeMethod not knowing what it's updating and so there not being an obvious class that could have this functionality -- TaskMetadata itself is hamstrung by it having to originally be a PropertySet, whereas if we were doign this now we'd have dedicated facilities in TaskMetadata for task execution metadata.

BrunoSanchez (approved):
This looks good to me.

PR #200, opened 2025-02-04
DM-48788: Switch configuration to ruff format and enable ruff isort checks

timj:
## Checklist

- [x] ran Jenkins


TallJimbo (approved):
I'm very happy about keeping `__all__` sorted.

timj:
The `__all__` sorting is RUFF022 but is only enabled with `--preview` so I haven't really enabled it.

PR #201, opened 2025-03-05
DM-46249-hotfix3 : Upload analysis_tools metrics from ap_verify so they can be retrieved in Chronograf

BrunoSanchez:
Add a new module, with argparsing tools. In particular class `AppendDict` to handle multiple optional keyword value inputs in cli scripts.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/__init__.py
+++ b/python/lsst/utils/__init__.py
@@ -11,6 +11,7 @@
 """General LSST Utilities."""
 
 from ._packaging import *
+from .argparsing import *
 from .deprecated import *
 from .doImport import *
 from .inheritDoc import *
--- a/python/lsst/utils/__init__.py
+++ b/python/lsst/utils/__init__.py
@@ -11,7 +11,6 @@
 """General LSST Utilities."""
 
 from ._packaging import *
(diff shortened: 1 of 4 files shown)


python/lsst/utils/__init__.py
```diff
@@ -11,6 +11,7 @@
 """General LSST Utilities."""
 
 from ._packaging import *
+from .argparsing import *
```
timj, line 14: Please do not export by default. We no longer do that in utils and these other exports are historical.

python/lsst/utils/argparsing.py
```diff
@@ -0,0 +1,79 @@
...
+__all__ = ["AppendDict"]
+
+import argparse
+import copy
+from collections.abc import Mapping
+from typing import Any
+
+
+class AppendDict(argparse.Action):
+    """An action analogous to the build-in 'append' that appends to a `dict`
```
timj, line 26:
```suggestion
    """An action analogous to the built-in 'append' that appends to a `dict`
```

ebellm (approved):

PR #202, opened 2025-03-24
SITCOM-1862: Create standard matplotlib style for System First Light construction paper

jeffcarlin:
The module can be import using:
`from lsst.utils.plotting import publication_plots`

To set the plot style (in python):
`publication_plots.set_rubin_plotstyle()`

To retrieve the dicts defining per-band colors, symbols, and linestyles:
`bands_dict = publication_plots.get_band_dicts()`

## Checklist

- [ ] ran Jenkins
- [ ] added a release note for user-visible changes to `doc/changes`

PR #205, opened 2025-03-26
DM-49728: Create standard matplotlib style for Rubin publications

jeffcarlin:
Supersedes #202 

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/plotting/publication_plots.py
+++ b/python/lsst/utils/plotting/publication_plots.py
@@ -10,28 +10,35 @@
 # license that can be found in the LICENSE file.
 """Utilities for making publication-quality figures."""
 
-import matplotlib.pyplot
-
-import lsst.utils.plotting
-
 __all__ = [
     "get_band_dicts",
     "set_rubin_plotstyle",
 ]
+
+from matplotlib import style
+
+from lsst.utils import getPackageDir
+from . import (
+    get_multiband_plot_colors,
+    get_multiband_plot_linestyles,
+    get_multiband_plot_symbols,
+)
 
 
 def set_rubin_plotstyle() -> None:
     """
     Set the matplotlib style for Rubin publications
     """
-    matplotlib.pyplot.style.use("lsst.utils.plotting.rubin")
-    print("Set up Rubin matplotlib plot style.")
+    utilsPath = getPackageDir("utils")
+    styleFile = utilsPath + "/python/lsst/utils/plotting/rubin.mplstyle"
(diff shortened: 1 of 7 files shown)


python/lsst/utils/plotting/publication_plots.py
```diff
@@ -0,0 +1,60 @@
...
+)
+
+
+def set_rubin_plotstyle() -> None:
+    """
+    Set the matplotlib style for Rubin publications
+    """
+    utilsPath = getPackageDir("utils")
+    styleFile = utilsPath + "/python/lsst/utils/plotting/rubin.mplstyle"
+    style.use(styleFile)
```
timj, line 35: This is documented to take `lsst.utils.plotting.x` form so I think what you had before should work. You can never use getPackageDir in PyPI packages (importlib.resources would work but seems like matplotlib is handling that for you).

jeffcarlin: I thought `import lsst.utils.plotting` was problematic, so was trying to avoid that here. Maybe I misunderstood.

timj: I'm talking about the file name here being replaced by a package resource. Like how you had it before.

My comment above about relative imports was to prevent the possibility of circular imports in the future.

python/lsst/utils/plotting/publication_plots.py
```diff
@@ -0,0 +1,56 @@
...
+# Use of this source code is governed by a 3-clause BSD-style
+# license that can be found in the LICENSE file.
+"""Utilities for making publication-quality figures."""
+
+__all__ = [
+    "get_band_dicts",
+    "set_rubin_plotstyle",
+]
+
+from matplotlib import style
```
timj, line 18: I've just realized that utils doesn't formally depend on matplotlib so I think you have to defer this import to be inside `set_rubin_plotstyle`. (`pip install lsst-utils` will not install matplotlib but `pip install lsst-utils[plotting]` will).

tests/test_matplotlib_figures.py
```diff
@@ -21,12 +21,16 @@
 
 import unittest
 
+from matplotlib import rcParams
```
timj, line 24: Please move this into the try block below so that the skip logic will work properly.

timj (approved):
Looks great. Thanks. One request: you need to move the matplotlib imports around a little to protect them in a world where matplotlib is not installed.

PR #206, opened 2025-03-29
DM-44862: Provide defaults for DbAuth constructor

andy-slac:
DbAuth required either the name of the file or envvar (or both)
for its constructor, but did not specify defaults. As a result
few packages had to define those to be the same in each package.
It is better to keep these values in one location and use them
as defautlts for constructor.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-44862.api.md
+++ b/doc/changes/DM-44862.api.md
@@ -1,0 +1,2 @@
+DbAuth constructor now provides default values for path and environment variable name.
+This avoids the need to specify the same values in each package that uses DbAuth.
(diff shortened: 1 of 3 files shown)


timj:
Package is broken on 3.13. There is ticket for that.

.pre-commit-config.yaml
```diff
@@ -10,7 +10,7 @@ repos:
       - id: check-toml
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.9.6
+    rev: v0.11.2
```
timj, line 13: formatting.yaml action is pinning ruff version to 0.9.x (to match rubinenv) so there is a possibility that you can get inconsistencies from moving version unless you also override the version in formatting.yaml (which is possible)

timj (approved):

PR #207, opened 2025-03-31
DM-34978: Allow disabling the timing decorator

parejkoj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/timer.py
+++ b/python/lsst/utils/timer.py
@@ -19,6 +19,7 @@ __all__ = ["duration_from_timeMethod", "logInfo", "profile", "timeMethod", "time
 import datetime
 import functools
 import logging
+import os
 import sys
 import time
 import traceback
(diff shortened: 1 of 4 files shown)


python/lsst/utils/timer.py
```diff
@@ -470,7 +487,7 @@ def profile(filename: str | None, log: LsstLoggers | None = None) -> Iterator[cP
         return
     from cProfile import Profile
 
-    profile = Profile()
+    profile = Profile(builtins=False)
```
timj, line 490: I don't think this is part of this ticket.

timj (approved):
Looks good. Thanks for persevering.

python/lsst/utils/timer.py
parejkoj: Ooops! I'd been so careful about keeping that out (it needs to be there to be able to use the profile decorator to test the timer decorator in profiling), per [DM-49470](https://rubinobs.atlassian.net/browse/DM-49470).

PR #209, opened 2025-04-03
DM-49860: Allow DbAuth to read creds from environment JSON string

timj:
@tcjennings does this do what you need?

@andy-slac we desperately need to switch to pydantic internally for validation...

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/utils/db_auth.py
+++ b/python/lsst/utils/db_auth.py
@@ -31,18 +31,6 @@ import yaml
 
 __all__ = ["DbAuth", "DbAuthError", "DbAuthPermissionsError"]
 
-DB_AUTH_ENVVAR = "LSST_DB_AUTH"
-"""Default name of the environmental variable that will be used to locate DB
-credentials configuration file. """
-
-DB_AUTH_PATH = "~/.lsst/db-auth.yaml"
-"""Default path at which it is expected that DB credentials are found."""
-
-DB_AUTH_CREDENTIALS_ENVVAR = "LSST_DB_AUTH_CREDENTIALS"
-"""Name of an environment variable containing the DB credentials directly
-in JSON format.
-"""
-
 
 class DbAuthError(RuntimeError):
     """Exception raised when a problem has occurred retrieving database
@@ -69,15 +57,16 @@ class DbAuth:
 
     Parameters
     ----------
-    path : `str` or None, optional
+    path : `str`, optional
         Path to configuration file, default path is ``~/.lsst/db-auth.yaml``.
-    envVar : `str` or None, optional
+    envVar : `str`, optional
         Name of environment variable pointing to configuration file, default is
-        ``LSST_DB_AUTH``.
-    authList : `list` [`dict`] or None, optional
+        ``LSST_DB_AUTH``. This is used if the environment variable is available
+        even if the path is specified.
+    authList : `list` [`dict`] or `None`, optional
         Authentication configuration. Used directly if given without referring
         to the environment.
-    credsEnvVar : `str` or None, optional
+    credsEnvVar : `str`, optional
         Name of environment variable containing the authentication
         configuration in JSON format. Default is ``LSST_DB_AUTH_CREDENTIALS``.
         If defined, this takes priority over reading credentials from file.
@@ -92,10 +81,10 @@ class DbAuth:
 
     def __init__(
         self,
-        path: str | None = None,
-        envVar: str | None = None,
+        path: str = "~/.lsst/db-auth.yaml",
+        envVar: str = "LSST_DB_AUTH",
         authList: list[dict[str, str]] | None = None,
-        credsEnvVar: str | None = None,
+        credsEnvVar: str = "LSST_DB_AUTH_CREDENTIALS",
     ):
         if authList is not None:
             self._db_auth_path = "<auth-list>"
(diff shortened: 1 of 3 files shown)


python/lsst/utils/db_auth.py
```diff
@@ -37,6 +38,11 @@
 DB_AUTH_PATH = "~/.lsst/db-auth.yaml"
 """Default path at which it is expected that DB credentials are found."""
 
+DB_AUTH_CREDENTIALS_ENVVAR = "LSST_DB_AUTH_CREDENTIALS"
+"""Name of an environment variable containing the DB credentials directly
+in JSON format
```
andy-slac, line 43:
```suggestion
in JSON format.
```

python/lsst/utils/db_auth.py
```diff
@@ -69,24 +75,44 @@ class DbAuth:
...
+        Name of environment variable containing the authentication
+        configuration in JSON format. Default is ``LSST_DB_AUTH_CREDENTIALS``.
+        If defined, this takes priority over reading credentials from file.
 
     Notes
     -----
-    At least one of ``path``, ``envVar``, or ``authList`` must be provided;
-    generally ``path`` should be provided as a default location.
+    At least one of ``path``, ``envVar``, ``credsEnvVar`` or ``authList`` must
+    be provided; generally ``path`` should be provided as a default location.
```
andy-slac, line 88: I missed that on my ticket, but with my changes all parameters are now optional, empty parameter list will give you a reasonable result.

andy-slac (approved):
Looks great.

andy-slac:
May deserve release note for API change. I agree that Pydantic would be nice.

python/lsst/utils/db_auth.py
```diff
@@ -69,24 +75,44 @@ class DbAuth:
...
     """
 
     def __init__(
         self,
         path: str | None = None,
         envVar: str | None = None,
         authList: list[dict[str, str]] | None = None,
+        credsEnvVar: str | None = None,
```
tcjennings, line 98: If the docstring describes a default value for this parameter, then that should be the default value for this parameter. There's no useful case where this should be `None`, and your implementation can be simplified because you need neither a module-level constant holding this value nor the extra test at L106.

python/lsst/utils/db_auth.py
```diff
@@ -69,24 +75,44 @@ class DbAuth:
...
     ):
         if authList is not None:
             self._db_auth_path = "<auth-list>"
             self.authList = authList
             return
+
+        # Credentials in JSON takes priority.
+        jsonEnvVar = credsEnvVar or DB_AUTH_CREDENTIALS_ENVVAR
+        jsonCreds = os.environ.get(jsonEnvVar, "")
```
tcjennings, line 107: Prefer the `os.getenv(credsEnvVar)` pattern, and let the default be `None` if the envvar doesn't exist.

python/lsst/utils/db_auth.py
timj: I rewrote it to say they are all optional.

python/lsst/utils/db_auth.py
```diff
@@ -69,24 +75,46 @@ class DbAuth:
...
+        if jsonCreds:
+            try:
+                self.authList = json.loads(jsonCreds)
+            except Exception as exc:
```
tcjennings, line 111: Prefer a `JSONDecodeError` here to the generic `Exception`.

python/lsst/utils/db_auth.py
```diff
@@ -69,24 +75,46 @@ class DbAuth:
...
+                raise DbAuthError(
+                    f"Unable to load DbAuth configuration using JSON environment variable {jsonEnvVar}"
+                ) from exc
+            self._db_auth_path = "<json-env-var>"
```
tcjennings, line 115: For whose benefit is this new attribute and token?

python/lsst/utils/db_auth.py
timj: I was following the existing scheme that @andy-slac introduced recently. I was wondering if he was using module-level constants to make it easier to change the default itself.

python/lsst/utils/db_auth.py
timj: This is not new. It's set to indicate the source of the auth information. It's returned in the `db_auth_path` property and if you look above you will see that there is a different value used if authList is given directly.

python/lsst/utils/db_auth.py
timj: I preferred `""` as the default for typing purposes so that `jsonCreds` is always a `str` and not `str | None`. I can change it since I think the `if` below will let mypy filter `None` out.

python/lsst/utils/db_auth.py
andy-slac: I think None as defaults were used in earlier versions of this code to detect a case when none of the parameters is provided. With my recent change it probably does not matter anymore, should be OK to use string defaults for all params.

timj: @tcjennings please take a look at the new version. If you could confirm on Jira  this meets your requirements that would be great. Thank you.

python/lsst/utils/db_auth.py
tcjennings: There are many weird tricks to convince mypy of The Thing. I don't love it, but the "narrow by first assertion" rule with a `if TYPE_CHECKING: assert __var__ is not None` avoids a lot of issues with "optional" types.

PR #211, opened 2025-04-09
DM-49951: Update publication plotting style

sr525:


python/lsst/utils/plotting/publication_plots.py
```diff
@@ -13,8 +13,20 @@
...
+    "stars_cmap",
+    "stars_color",
+    "galaxies_cmap",
+    "galaxies_color",
+    "accent_color",
 ]
 
+import numpy as np
+import seaborn as sns
+from matplotlib import cm, colors
```
timj, line 27: matplotlib is an optional dependency of utils, which is why it was a protected import in set_rubin_plotstyle below.

seaborn is definitely not a dependency of lsst-utils and has to be optional. It also needs to be added to the optional plotting section in pyproject.toml.

timj: (handling these optional dependencies is why I was advocating a lsst-plotutils package in the beginning).

leannep (approved):

PR #212, opened 2025-04-25
DM-50490: Return duration in time_this context manager

timj:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/utils/timer.py
+++ b/python/lsst/utils/timer.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 __all__ = ["duration_from_timeMethod", "logInfo", "profile", "timeMethod", "time_this"]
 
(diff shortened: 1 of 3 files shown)


python/lsst/utils/timer.py
```diff
@@ -338,6 +339,14 @@ def pass_through(func: Callable) -> Callable:
         return decorator_timer(_func)
 
 
+@dataclasses.dataclass
+class _TimerResult:
+    duration: float | None = None
+    mem_current_usage: float | None = None
```
timj, line 345: The problem with the memory numbers is that we only calculate them if the log message was going to deliver them, so we likely need an extra parameter for time_this to force computation even if the log message is not being shown.

TallJimbo (approved):

PR #213, opened 2025-05-09
SP-2200: Consider updates to the filter linestyle advice with matplotlib

plazas:
## Checklist

- [ ] ran Jenkins


python/lsst/utils/plotting/figures.py
```diff
@@ -139,10 +139,10 @@ def get_multiband_plot_linestyles() -> dict:
...
-        "g": ":",
-        "r": "-",
-        "i": "-.",
-        "z": "--",
+        "g": "(0, (3, 1, 1, 1))",
+        "r": "-.",
+        "i": "-",
+        "z": "(0, (3, 1, 1, 1, 1, 1))",
         "y": ":",
     }
```
jeffcarlin, line 148: I wonder if we should add a warning here that the linestyles for g and z bands may cause a matplotlib error, with a suggested workaround until the matplotlib fix makes it to rubin-env?

jeffcarlin (changes requested):
A couple suggestions we should fix, and also make sure to run Jenkins after fixing the errors on the PR checks.

python/lsst/utils/plotting/figures.py
plazas: I did have a comment after the code. I just changed it a bit based on your comment above, but the original comment I made explained the workaround.

plazas: Let me know if it looks good to you!

PR #214, opened 2025-05-16
DM-50926: Change default color cycle in plotting utils

jeffcarlin:
Changed the default color cycle to that from seaborn-v0_8-colorblind colormap. See [https://seaborn.pydata.org/tutorial/color_palettes.html](https://seaborn.pydata.org/tutorial/color_palettes.html) for illustrations of this colormap.


sr525:
This colour palette is much more appealing than the previous option.

sr525 (approved):

PR #225, opened 2025-08-14
DM-51290: Fix type annotations for timeMethod.

TallJimbo:
Using this as a decorator no longer obliterates the annotations on the decorated function.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/utils/timer.py
+++ b/python/lsst/utils/timer.py
@@ -27,7 +27,7 @@ import time
 import traceback
 from collections.abc import Callable, Collection, Iterable, Iterator, MutableMapping
 from contextlib import contextmanager, suppress
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, overload
 
 from astropy import units as u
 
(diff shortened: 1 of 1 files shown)


timj:
If you are changing the base python requirement you also need to change pyproject.toml

python/lsst/utils/timer.py
```diff
@@ -233,6 +233,22 @@ def logInfo(
...
+def timeMethod(_func: _F) -> _F: ...
+
+
+@overload
+def timeMethod(
+    *,
+    metadata: MutableMapping | None = None,
+    logger: LsstLoggers | None = None,
+    logLevel: int = logging.DEBUG,
+) -> Callable[[_F], _F]: ...
```
MichelleGower, line 249: Does there need to be a description of the returned Callable in the docstring?

MichelleGower (approved):
A docstring question.  Merge approved.

python/lsst/utils/timer.py
TallJimbo: The returned callable is actually the decorator itself, so I think the original docs still cover it.

PR #227, opened 2025-09-15
DM-47313: Try to work around change in behavior in get_referrers for python 3.13

timj:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/utils/introspection.py
+++ b/python/lsst/utils/introspection.py
@@ -443,6 +443,10 @@ def _get_clean_refs(objects: list) -> list:
     """
     refs = gc.get_referrers(*objects)
     refs.remove(objects)
+    # On Python 3.13 gc.get_referrers returns the original objects in a list
+    # and in a tuple. Remove both if present.
+    if (tobjects := tuple(objects)) in refs:
+        refs.remove(tobjects)
     iterators = [x for x in refs if type(x).__name__.endswith("_iterator")]
     for i in iterators:
         refs.remove(i)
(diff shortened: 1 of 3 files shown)


python/lsst/utils/introspection.py
```diff
@@ -443,6 +443,10 @@ def _get_clean_refs(objects: list) -> list:
     """
     refs = gc.get_referrers(*objects)
     refs.remove(objects)
+    # On Python 3.13 gc.get_referrers returns the original objects in a list
+    # and in a tuple. Remove both if present.
+    if (tobjects := tuple(objects)) in refs:
+        refs.remove(tobjects)
```
andy-slac, line 449: In pre-3.13 if there is an actual tuple with `objects`, will it be removed too? If we need something more reliable we probably need to check `id()` of the referrers.

andy-slac (approved):
Looks good but maybe it removes too much in pre-3.13 Python.

python/lsst/utils/introspection.py
timj: Yes, it would remove them in python 3.12 as well. I'm not sure how much we care since this is a debugging tool. I'm not entirely sure where the extra tuple is coming from. I could do a version check here as well if you prefer that.

andy-slac: Extra tuple is probably from `*objects` when they are passed to `get_referrers`. I think something like this could work for all versions:

    objects_tuple = tuple(objects)
    refs = gc.get_referrers(*objects_tuple)
    ids_to_drop = {id(objects), id(objects_tuple)}
    refs = [ref for ref in refs if id(ref) not in ids_to_drop]
    refs = [ref for ref in refs if not type(x).__name__.endswith("_iterator")]
    return refs

timj: You win the prize! 🎉

PR #237, opened 2026-01-20
DM-53868: Switch to new sphinxutils + documenteer2 doc build

timj:
Requires lsst/sconsUtils#140

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -2,28 +2,11 @@
 
 This configuration only affects single-package Sphinx documenation builds.
 """
+# ruff: noqa: F403, F405
 
-from documenteer.conf.pipelinespkg import *  # noqa: F403, import *
+from documenteer.conf.guide import *
 
-project = "utils"
-html_theme_options["logotext"] = project  # noqa: F405, unknown name
-html_title = project
-html_short_title = project
-doxylink = {}
-exclude_patterns = ["changes/*"]
-nitpick_ignore_regex = [
-    ("py:.*", r"lsst\..*"),  # Ignore warnings from links to other lsst packages.
-    ("py:.*", "cProfile.Profile"),  # Link does not resolve to py3 docs.
-    ("py:class", "a set-like.*"),  # collections.abc.Mapping inheritance.
-    ("py:class", "a shallow copy.*"),  # collections.abc.Mapping inheritance.
-    ("py:class", r"None\.  .*"),  # collections.abc.Mapping inheritance.
-    ("py:class", "an object providing.*"),  # collections.abc.Mapping inheritance.
-]
-nitpick_ignore = [
-    ("py:class", "module"),  # Sphinx has problems with types.ModuleType.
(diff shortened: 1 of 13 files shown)


doc/conf.py
```diff
@@ -2,28 +2,11 @@
...
-    ("py:class", "module"),  # Sphinx has problems with types.ModuleType.
-    ("py:obj", "logging.DEBUG"),  # Python 3 constant can not be found.
-    ("py:class", "unittest.case.TestCase"),  # Sphinx can not see TestCase.
-    ("py:obj", "deprecated.sphinx.deprecated"),  # No intersphinx for deprecated package.
-    ("py:class", "LsstLoggers"),  # Sphinx does not understand type aliases
-    ("py:obj", "structlog"),  # structlog does not have intersphinx mapping.
-]
+exclude_patterns.append("changes/*")
+
+intersphinx_mapping["astropy"] = ("http://docs.astropy.org/en/stable", None)
```
timj: @jonathansick I couldn't work out how to do this correctly in the documenteer.toml.

Also what was the motivation for dropping even the python docs from the defaults?

python/lsst/utils/deprecated.py
```diff
@@ -35,19 +35,19 @@ def deprecate_pybind11(obj: Any, reason: str, version: str, category: type[Warni
 
     Parameters
     ----------
-    obj : function, method, or class
+    obj : `~collections.abc.Callable`
```
TallJimbo, line 38: While it's an improvement in a lot of cases here, this new rigidity does make me sad (I  like "convertible to `ResourcePath`" as a type, for example).  Not much we can do about it, I imagine.

TallJimbo (commented):
Changes seem fine, but they're not working for me locally; I get
```
$ package-docs build
Running Sphinx v8.1.3

Configuration error:
sphinx.show_github_edit_link is True by the project.github_url is not set.
```
I suspect "by" should be "but" in that error message.

I've tried setting `show_github_edit_link = [Ff]alse` in both `conf.py` and `documenteer.toml` to no avail.

python/lsst/utils/deprecated.py
timj: I did add `array-like` to the ignore list because fixing all those seemed a step too far.

TallJimbo: If there's an ignore list we can add our own exceptions to then I'm totally happy with this change.  I'm all for better guards against non-types due to mispellings, as long as we can carve out exceptions.

TallJimbo (approved):
As @timj pointed out offline, my `show_github_edit_link` issue was just due to not having rebuild the package with the `sconsUtils` branch.

python/lsst/utils/deprecated.py
timj: Yes. It's in the documenteer.toml file.

doc/conf.py
timj: I'm supposed to use `[sphinx.intersphinx.projects]` and that works.

PR #243, opened 2026-07-07
Add DeprecatedDict mapping

fred3m:
DeprecatedDict is a dict that warns the first time a deprecated key is read or written through targeted single-key access, while leaving bulk operations (copy, iterate, serialize, merge) silent. It eases migrating a field out of a mapping and onto a typed attribute while keeping the old key readable during a deprecation period.

### DIFF
--- a/python/lsst/utils/deprecated.py
+++ b/python/lsst/utils/deprecated.py
@@ -11,12 +11,13 @@
 
 from __future__ import annotations
 
-__all__ = ["deprecate_pybind11", "suppress_deprecations"]
+__all__ = ["DeprecatedDict", "deprecate_pybind11", "suppress_deprecations"]
 
+import copy
 import functools
 import unittest.mock
 import warnings
-from collections.abc import Iterator
+from collections.abc import Iterator, Mapping
 from contextlib import contextmanager
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


python/lsst/utils/deprecated.py
```diff
@@ -95,3 +96,160 @@ def suppress_deprecations(category: type[Warning] = FutureWarning) -> Iterator[N
         warnings.simplefilter("ignore", category)
         with unittest.mock.patch.object(warnings, "simplefilter"):
             yield
+
+
+class DeprecatedDict(dict):
+    """A `dict` that warns when deprecated keys are used.
```
taranu: are first used

python/lsst/utils/deprecated.py
```diff
@@ -95,3 +96,160 @@ def suppress_deprecations(category: type[Warning] = FutureWarning) -> Iterator[N
...
+
+    ``DeprecatedDict`` behaves exactly like a built-in `dict`, except that
```
taranu: This whole paragraph is basically redundant. Consider instead merging with the next paragraph e.g.:

``DeprecatedDict`` behaves exactly like a built-in `dict`, except that it keeps a set of deprecated keys. The first time each deprecated key is accessed through a single-key operation (``d[key]``, `~dict.get`, `~dict.pop`, `~dict.setdefault`, ``d[key] = value`` and ``del d[key]``), it issues a deprecation warning.

python/lsst/utils/deprecated.py
```diff
@@ -95,3 +96,160 @@ def suppress_deprecations(category: type[Warning] = FutureWarning) -> Iterator[N
...
+    The methods that warn are the ones that reach for one specific key:
+    ``d[key]``, `~dict.get`, `~dict.pop`, `~dict.setdefault`,
+    ``d[key] = value`` and ``del d[key]``. Bulk operations that copy or
+    walk the whole mapping -- iteration, ``dict(d)``, ``{**d}``,
+    `~dict.keys`, `~dict.values`, `~dict.items` and `~dict.update` -- never
+    warn, so the container can be serialized, copied, or merged without
+    spurious noise. Membership tests (``key in d``) do not warn either.
+
+    This is a transitional helper for moving a field out of a mapping and
+    onto a typed attribute while keeping the old key readable for a
```
taranu: Well, that's what you made it for, but surely you could use it for more general reasons (e.g. just removing a field entirely). So I would describe this more as one possible/suggested use case.

python/lsst/utils/deprecated.py
```diff
@@ -95,3 +96,160 @@ def suppress_deprecations(category: type[Warning] = FutureWarning) -> Iterator[N
...
+    deprecation period. Because deprecated keys remain live entries, this
+    is not an access barrier -- it only announces that the key is going
```
taranu: Again this is redundant and basically describing what deprecation is.

python/lsst/utils/deprecated.py
```diff
@@ -95,3 +96,160 @@ def suppress_deprecations(category: type[Warning] = FutureWarning) -> Iterator[N
...
+        *args: Any,
+        deprecated_keys: Iterable[Any] = (),
+        reason: str = "",
+        version: str = "",
+        category: type[Warning] = FutureWarning,
+        stacklevel: int = 2,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self._deprecated_keys: set[Any] = set(deprecated_keys)
```
taranu: Should this check that the inputs are a set through e.g `len(list(deprecated_keys)) == len(set(deprecated_keys))`?

python/lsst/utils/deprecated.py
```diff
@@ -95,3 +96,160 @@ def suppress_deprecations(category: type[Warning] = FutureWarning) -> Iterator[N
...
+        internals.
+    **kwargs : `~typing.Any`
+        Keyword arguments forwarded to `dict`, used to populate the
+        initial contents.
+    """
+
+    def __init__(
+        self,
+        *args: Any,
+        deprecated_keys: Iterable[Any] = (),
```
taranu: Wouldn't it make more sense to have deprecated be a dict of keys and reasons? Having the same reason for every key seems limiting, even for your use case.

tests/test_deprecated.py
```diff
@@ -19,10 +19,107 @@
...
 import lsst.utils.tests
+from lsst.utils.deprecated import DeprecatedDict
+
+
+class DeprecatedDictTestCase(lsst.utils.tests.TestCase):
+    """Test `~lsst.utils.deprecated.DeprecatedDict`."""
+
+    def makeDict(self, **kwargs):
+        # A DeprecatedDict with one deprecated key ("old") and one live
+        # key ("new"), built inside a filter so construction never leaks
```
taranu: This should never warn since deprecated_keys is provided; why is catching necessary?

tests/test_deprecated.py
```diff
@@ -19,10 +19,107 @@
...
+            warnings.simplefilter("ignore")
+            _ = d["old"]  # Latch the original.
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            clone = copy.deepcopy(d)
+            # The clone inherits the "already warned" latch, so it too is
+            # silent for the key already warned on the original.
+            _ = clone["old"]
+        self.assertIsInstance(clone, DeprecatedDict)
+        self.assertEqual(dict(clone), {"old": 1, "new": 2})
```
taranu, line 133: Consider adding e.g. `self.assertEqual(dict(clone), dict(d))` to ensure that d isn't modified either.

python/lsst/utils/deprecated.py
fred3m: Yeah, that's a good idea.

python/lsst/utils/deprecated.py
fred3m: This is unnecessary now that deprecated_keys is a dict of keys to reasons, based on your other suggestion.

tests/test_deprecated.py
fred3m: Good catch I was being too cautious

timj:
@fred3m I'm confused as to how you deleted the PR template for this repo that was asking you to make sure you had included a news fragment for the release notes...

fred3m:
Sorry, I didn't think that this was a release notes worthy addition.

timj:
For utils a whole new API is definitely release notes worthy 😄

timj:
I'll try to remember to add a note about it next week when I make the v30 release.

PR #245, opened 2026-07-15
DM-55482: Improve coverage of disable_implicit_threading

timj:
Add NUMBA_NUM_THREADS, ARROW_IO_THREADS, VECLIB_MAXIMUM_THREADS, RAYON_NUM_THREADS, POLARS_MAX_THREADS, and BLIS_NUM_THREADS to the environment variables set by set_thread_envvars.

Resize existing pyarrow thread pools explicitly since they are sized from the environment only when each pool is first used and do not react to environment variables after creation.

Log a warning when threadpoolctl is not installed since thread pools of already-loaded libraries can then no longer be limited.

## Checklist

- [x] ran Jenkins


erykoff (approved):
This is all good; I wonder if we want something here for `treecorr` as well (checking if it's been loaded).  Though this is a known issue in treecorr and will be fixed hopefully soon so it's probably fine for us to patch that where we actually use `treecorr`.

timj:
@erykoff there needed to be a slight tweak to the numba handling since `@njit` rechecks the environment variable and complains if there is an inconsistency with how it was first set up.

PR #247, opened 2026-07-20
DM-55547: Add perf-NNN.map files to the leak test ignore list

timj:
These can be generated by the perf-profiling infrastructure in Python, especially with JIT compilation.

## Checklist

- [x] ran Jenkins


roceb (approved):
Easy fix for a weird bug

